PR #4, opened 2017-10-13
Fixing examples

kanaadp:
Fixed all of the example scenarios provided (some were not compatible with params classes).


cathywu:
LGTM

PR #24, opened 2018-08-22
modified the rllab setup

AboudyKreidieh:
they should work now for everyone that had isssue

### DIFF
--- a/docs/source/flow_setup.rst
+++ b/docs/source/flow_setup.rst
@@ -120,11 +120,9 @@ to install the `rllab-multiagent` library, follow the below instructions
     git clone https://github.com/cathywu/rllab-multiagent.git
     cd rllab-multiagent
     conda env create -f environment.yml
-    source activate flow-rllab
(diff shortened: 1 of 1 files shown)


docs/source/flow_setup.rst
```diff
@@ -152,7 +150,7 @@ Getting started (rllab-multiagent)
 To run any of the RL examples, make sure to run
 ::
 
-    source activate flow
+    source activate flow-rllab
```
eugenevinitsky: This is not the case if you're not using rllab; you won't have this env.

AboudyKreidieh: u create the environment during the rllab setup instructions

AboudyKreidieh: this is in the Setting up section for rllab btw, which is totally optional and assumes u've installed rllab

eugenevinitsky: Mmm, I was confused by the git diff. Okay!

PR #28, opened 2018-08-23
Amend setup

eugenevinitsky:
- Created an environment.yml file that supersedes the rllab-multi-agent environment.yml file
- updated the setup instructions to handle it


eugenevinitsky:
@AboudyKreidieh this I think is a time-sensitive PR (due to homework0)

PR #38, opened 2018-08-24
Yapf

eugenevinitsky:
Added yapf changes to make the code look a bit nicer. All changes are entirely cosmetic, nothing else.


eugenevinitsky:
I'm sorry for whoever has to code review this.

AboudyKreidieh:
also, people should merge right away or this'll be a nightmare later

eugenevinitsky:
Yeah that’s the logic. A strict style guide that we can enforce with one line of code will be useful as we grow. Not sure I love every aspect of it either; we can create a style file for yapf if you have preferences.

PR #39, opened 2018-08-25
Setup fix

eugenevinitsky:
- The links to conda and miniconda were broken 
- Makefile for docs was missing


nskh:
This commit includes all of the YAPF changes - so are we waiting on #38 before merging this in? Would be best if it was separate, right?

PR #41, opened 2018-08-26
removed positions and lanes from initial_config

AboudyKreidieh:
- removed positions and lanes from `initial_config`. These attributes are modified by the generator class, and causes confusion and some bugs when trying to use the same parameters on a new task (e.g. what we do in the homework, which is all written in jupyter). This removes the need to write:
```
initial_config.lanes, initial_config.positions = None, None
```
- minor modifications to the docstrings

### DIFF
--- a/flow/core/generator.py
+++ b/flow/core/generator.py
@@ -33,9 +33,9 @@ class Generator(Serializable):
 
         Attributes
         ----------
-        net_params: NetParams type
+        net_params : NetParams type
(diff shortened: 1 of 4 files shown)


eugenevinitsky:
It doesn't look like you've removed initial_positions and lanes from the actual initialconfig class?

AboudyKreidieh:
see flow/core/params.py. they're called `positions` and `lanes`

PR #46, opened 2018-08-26
cleanup to docstrings in tests

AboudyKreidieh:


tests/setup_scripts.py
```diff
@@ -1,3 +1,9 @@
+"""Specifies several methods for creating scenarios and environments.
```
eugenevinitsky: Does this s pass the docs @AboudyKreidieh ?

PR #49, opened 2018-08-27
cleaned up docstring in envs

AboudyKreidieh:


flow/envs/bay_bridge/base.py
```diff
@@ -209,12 +209,15 @@ def apply_toll_bridge_control(self):
 
     # TODO: decide on a good reward function
     def compute_reward(self, state, rl_actions, **kwargs):
+        """See class definition."""
         return np.mean(self.vehicles.get_speed(self.vehicles.get_ids()))
 
     """ The below methods need to be updated by child classes. """
 
     def _apply_rl_actions(self, rl_actions):
+        """Update by child classes."""
```
nskh: Should this read "update using child classes" or "implemented by child classes"?

flow/envs/green_wave_env.py
```diff
@@ -468,8 +473,10 @@ def __init__(self, env_params, sumo_params, scenario):
     @property
     def observation_space(self):
         """
-        Partial observed state space. Velocities, distance to intersections,
-        edge number (for nearby vehicles) traffic light state
+        Partial observed state space.
```
nskh: Super nitpicky, but should this be "Partially observed"?

eugenevinitsky: Not too nitpicky, this is a good catch.

PR #53, opened 2018-08-27
docstring cleanup for flow/core/

AboudyKreidieh:


flow/core/experiment.py
```diff
@@ -6,9 +8,47 @@
...
+
+        >>> from flow.core.params import SumoParams
+        >>> sumo_params = SumoParams(emission_path="./data")
+
+    Once you have included this in your environment, run your SumoExperiment
+    object as follows:
+
+        >>> exp.run(num_runs=1, num_steps=1000, convert_to_csv=True)
+
+    After the experiment is complete, look at the "./data" directory. There
```
eugenevinitsky: Nice, props.

flow/core/params.py
```diff
@@ -522,7 +549,7 @@ def add(self,
             probability=None,
             number=None,
             **kwargs):
-        """Specifies a new inflow for a given type of vehicles and edge.
+        r"""Specify a new inflow for a given type of vehicles and edge.
```
eugenevinitsky: Don't think this r is intended?

AboudyKreidieh: docstring standards is complaining about this

PR #56, opened 2018-08-27
docstring cleanup for scenarios

AboudyKreidieh:


flow/scenarios/__init__.py
```diff
@@ -25,20 +27,12 @@
...
-    "CircleGenerator", "MergeGenerator", "NetFileGenerator",
-    "TwoLoopOneMergingGenerator"
-]
-
-# custom scenarios
-__all__ += [
-    "BayBridgeScenario", "BayBridgeTollScenario", "BottleneckScenario",
-    "Figure8Scenario", "SimpleGridScenario", "HighwayScenario", "LoopScenario",
-    "MergeScenario", "NetFileScenario", "TwoLoopsOneMergingScenario"
+__all__ = [
```
eugenevinitsky: Why do this? This seems less clean.

flow/scenarios/bay_bridge_toll/gen.py
```diff
@@ -1,13 +1,14 @@
+"""Contains the Bay Bridge toll generator class."""
+
 from flow.scenarios.netfile.gen import NetFileGenerator
 
 
 class BayBridgeTollGenerator(NetFileGenerator):
-    """
-    Bay Bridge bottleneck generator.
-    """
+    """Bay Bridge bottleneck generator."""
```
eugenevinitsky: seems misnamed given the above information?

flow/scenarios/bottleneck/gen.py
```diff
@@ -1,17 +1,17 @@
...
+
 from flow.core.generator import Generator
 import numpy as np
 
 
 class BottleneckGenerator(Generator):
-    """
-    Generator class for simulating the Bay Bridge toll. No parameters needed
-    from net_params (the network is not parametrized)
+    """Generator class for simulating the Bay Bridge toll.
```
eugenevinitsky: Seems misnamed given the above naming scheme

flow/scenarios/loop/gen.py
```diff
@@ -1,27 +1,23 @@
...
 from flow.core.generator import Generator
 
 from numpy import pi, sin, cos, linspace
 
 
 class CircleGenerator(Generator):
-    """
-    Generator for loop circle used in MIT traffic simulation.
-    """
+    """Generator for loop circle used in MIT traffic simulation."""
```
eugenevinitsky: Remove MIT?

flow/scenarios/loop_merge/gen.py
```diff
@@ -1,18 +1,18 @@
...
 from numpy import pi, sin, cos, linspace
 
 
 class TwoLoopOneMergingGenerator(Generator):
-    """
-    Generator for a two-loop network in which both loops merge into a common
-    lane.
+    """Generator for a two loop merge network.
+
+    This network conssits of two loops that both merge into a common lane.
```
eugenevinitsky: Consists

flow/scenarios/__init__.py
AboudyKreidieh: Docstring conventions don't like it when we do this:
```
WARNING: Error in file flow/scenarios/__init__.py: Could not evaluate contents of __all__. 
That means pydocstyle cannot decide which definitions are
public. Variable __all__ should be present at most once in
each file, in form
```

eugenevinitsky: kk, can't do much about that

AboudyKreidieh:
Thanks for the thorough review.

PR #75, opened 2018-08-31
Leaderboard

fywu85:
Fix minor bugs in previous PR. This one is tested with and passed `docker build`.


flow/utils/leaderboard/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+	apt-get -y upgrade && \
+	apt-get install -y \
+    vim \
+    apt-utils
+
+# Flow
+RUN cd ~ && \
+	git clone https://github.com/flow-project/flow.git && \
+    cd flow && \
+    git checkout leaderboard && \
```
nskh: If we're merging this into master, is the `git checkout leaderboard` going to be necessary?

fywu85: My plan is to keep the leaderboard branch as sort of devel branch and only push stable, tested codes to master once in a while. I just realized that I failed all the checks, again... Let me fix this!

AboudyKreidieh: switch to master

flow/utils/leaderboard/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+	libxml2-dev \
+	libxslt1-dev \
+	openjdk-8-jdk
+
+# SUMO
+RUN cd ~ && \
+	git clone --recursive https://github.com/eclipse/sumo.git && \
+	cd sumo && \
+	#git checkout b6f37dd && \
+	git checkout cbe5b73 && \
```
AboudyKreidieh: can u fix to `016c09d306`?

fywu85:
@AboudyKreidieh This should be good to go.

PR #77, opened 2018-09-01
Include binary sumo via wheel

pcmoritz:
This should make it much easier to install flow by removing the "compile SUMO" step in the installation instructions.

The wheel can be built in the following way (this only needs to be done once and then the resulting wheel can be distributed for easy installation):

- First build sumo
- Drop the sumo and sumo-gui binaries into flow/flow/bin/
- Run the following in the flow (root) directory:
```
python setup.py bdist_wheel
```

The resulting wheel will be in `flow/dist/<wheelname>.whl`. It can be installed via
```
pip install <wheelname>.whl
```

For easier installation and automatically figuring out the distribution, the wheel should be uploaded to pypy.

### DIFF
--- a/flow/envs/base_env.py
+++ b/flow/envs/base_env.py
@@ -178,9 +178,12 @@ class Env(gym.Env, Serializable):
                         time.sleep(1.0 * int(time_stamp[-6:]) / 1e6)
                         port = sumolib.miscutils.getFreeSocketPort()
 
+                script_dir = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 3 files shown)


pcmoritz:
I also added a little script that builds the wheels for all python versions on macOS and attached them here.

[mac_os_wheels.zip](https://github.com/flow-project/flow/files/2342049/mac_os_wheels.zip)

AboudyKreidieh:
we've added this since (so stale). but thanks @pcmoritz for the suggestion!

PR #62, opened 2018-08-28
New way to install flow

lucfisc:
Add a Dockerfile that can run with a desktop on your browser with every flow packages installed and ready to use

PR #80, opened 2018-09-03
Sumo wheel

AboudyKreidieh:
- created a wheel for sumo tools (traci, etc..) and separate prebuilt binaries for ubuntu 14.04 and 16.04+. In this PR, I included them to the travis setup instructions
- once #62 is up (just to avoid merge conflicts), I'll create a separate PR to include them into the setup instructions. 
- this significantly simplifies and speeds up the installation procedure, but NEEDS to be thoroughly vetted before we include it (ideally before September 6)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ before_install:
   -         rm binaries-ubuntu1404.tar.xz
   -         chmod +x *
   -         export HOME="$HOME:$(pwd)"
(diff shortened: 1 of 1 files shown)


eugenevinitsky:
Is this stale?

AboudyKreidieh:
this is me adding it to travis. needs to be updated

AboudyKreidieh:
actually, moving this into a new PR to add osx travis builds as well

PR #85, opened 2018-09-04
Install sumo binaries

AboudyKreidieh:
- modified the setup instructions to install sumo binaries into flow/bin. The binaries were built on my and Eugene's machines, as well as on AWS instances.
- sumo tools (e.g. traci) are installed in `setup.py` via a wheel. The wheel works for all operating systems (as far as I can tell so far)
- we need many people to test that the binaries work for them before this is merged. There may be some dependencies that won't work, and we can only really find out by testing this multiple times and finding the discrepancies


eugenevinitsky:
Can we temporarily flip the two, i.e. suggest the binaries if things don't work with the normal sumo install?

eugenevinitsky:
This seems like a fair compromise between safety and having these up

eugenevinitsky:
Thoughts?

AboudyKreidieh:
In my mind, we present them with the quick and easy way, and if that doesn't work they can put the work in and get some by themselves. I'm open to flipping it though.

docs/source/flow_setup.rst
```diff
@@ -84,66 +89,80 @@ Finally, test your sumo install and version by running the following commands
...
+
+In order to install everything you will need from SUMO, run one of the below
+scripts from the Flow main directory. Choose the script that matches the
+operating system you are running.
+
+For Ubuntu 14.04:
+
+::
+
+    scripts/setup_sumo_ubuntu1404.sh
```
eugenevinitsky: this is a shell script, do they need to chmod it first?

lucfisc:
@AboudyKreidieh @eugenevinitsky 
It works on my ubuntu 16:04 docker

docs/source/flow_setup.rst
AboudyKreidieh: haven't had to so far. this might be because we start the files with
```
#bin/bash
```

eugenevinitsky: ah nice

Issue #95, opened 2018-09-05
[readthedocs] Introduction


cathywu:
Introduction page should be more descriptive and include the following:

- What is Flow?
- RL loop diagram
- Architecture diagrams
- Some background on traffic flow

Other suggestions?

eugenevinitsky:
Probably a description fo the problems Flow solves relative to other tools/what it's really good for

kjang96:
Can this be closed?

PR #105, opened 2018-09-05
More descriptive Flow introduction text

cathywu:
- Added a description "Why Flow?" (addresses part of #95)
- Removed outdated team list, updated credits


cathywu:
I agree with you @nskh, let's handle that in a separate PR since we will want to put alumni somewhere too. Let's view that part of this PR as just removing redundant team information.

PR #150, opened 2018-09-17
resolved #116

### LINKED ISSUE #116: Bug in multi lane headways utility function in flow/core/vehicles.py
When computing the leaders on a multi lane, sometimes vehicles are left out.  If there is just one vehicle in front of the rl_vehicle on another lane, the index would be zero and therefore the condition in line 1159 is not satisfied, although a leader was found. 
I suggest removing the -1 in this line and check below, if there is at least one vehicle behind the rl_vehicle when checking the rl_vehicles lane.

AboudyKreidieh:
resolved #116

### DIFF
--- a/flow/core/vehicles.py
+++ b/flow/core/vehicles.py
@@ -1156,7 +1156,7 @@ class Vehicles:
 
                 # if you are at the end or the front of the edge, the lane
                 # leader is in the edges in front of you
-                if index < len(positions) - 1:
+                if index < len(positions):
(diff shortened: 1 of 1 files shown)


eugenevinitsky:
@AboudyKreidieh can you merge this into the PR I pushed; your solution is cleaner than mine but I've added a ton of tests that you should make sure your solution passes

AboudyKreidieh:
this is stale

PR #153, opened 2018-09-17
Multi lane bug fix

### LINKED ISSUE #147: Write more precise tests for _multi_lane_headways_util
As pointed out in #116, the _multi_lane_headways function will miss vehicles in adjacent lanes if there's only one vehicle in that lane. There should be tests that cover this as well as obviously the aforementioned bug fix

eugenevinitsky:
Added tests to _multi_lane_headways_util 
Fixed _multi_lane_headways_util to pass the tests 
Modified gen_custom_start_pos to allow for passing of explicit vehicle positions 
Fix #147

### DIFF
--- a/flow/scenarios/highway/gen.py
+++ b/flow/scenarios/highway/gen.py
@@ -71,7 +71,7 @@ class HighwayGenerator(Generator):
 
     def specify_routes(self, net_params):
         """See parent class."""
-        num_edges = net_params.additional_params.get("length", 1)
+        num_edges = net_params.additional_params.get("num_edges", 1)
         rts = {}
         for i in range(num_edges):
             rts["highway_{}".format(i)] = ["highway_{}".format(j) for j in range(i, num_edges)]
--- a/flow/scenarios/highway/gen.py
+++ b/flow/scenarios/highway/gen.py
@@ -25,15 +25,11 @@ class HighwayGenerator(Generator):
         segment_lengths = np.linspace(0, length, num_edges+1)
 
         nodes = []
-        for i in range(num_edges):
+        for i in range(num_edges+1):
             nodes += [{
-                "id": "begin_{}".format(i),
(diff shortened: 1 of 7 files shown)


eugenevinitsky:
This should fix #116

flow/scenarios/highway/gen.py
```diff
@@ -20,30 +21,34 @@ def __init__(self, net_params, base):
...
-            "type": "highwayType",
-            "from": "begin",
-            "to": "end",
-            "length": repr(length)
-        }]
+        num_edges = net_params.additional_params.get("num_edges", 1)
+        segment_length = length/float(num_edges)
+
+        edges = []
+        for i in range(num_edges):
```
AboudyKreidieh: sort of at a high level, why do we want to cut the highway into sections?

flow/scenarios/highway/scenario.py
```diff
@@ -39,11 +41,36 @@ def __init__(self,
...
+            config parameters
+
+        Returns
+        -------
+        startpositions : list of tuple (float, float)
+            list of start positions [(edge0, pos0), (edge1, pos1), ...]
+        startlanes : list of int
+            list of start lanes
+        """
+        return kwargs["start_positions"], kwargs["start_lanes"]
```
AboudyKreidieh: do we want this for all scenarios? if so, would be a good idea to create an issue about this

flow/scenarios/highway/gen.py
eugenevinitsky: This lets me write tests for the methods that get leaders for subsequent edges

flow/scenarios/highway/scenario.py
eugenevinitsky: It seems like a one-off at the moment? I don't see a compelling case for carefully specifying the initial position of vehicles but happy to be convinced otherwise

PR #154, opened 2018-09-17
DO NOT MERGE: New routes class

kjang96:
I created a new Routes class for simplification and to make route distribution a possibility (vehicles can take different routes starting from the same origin with some probability). This involves changes in a few different places outside of the new routes.py file: mostly generator.py and base_env.py. 

As an example, the routes of sumo/sugiyama.py are updated to reflect this change.

**I'd like some feedback on this abstraction before refining it.**


eugenevinitsky:
Adding @AboudyKreidieh for additional comments

AboudyKreidieh:
This is stale, and a new router PR is in the works, so closing

PR #160, opened 2018-09-19
Addressing issue 145

### LINKED ISSUE #145: Move unexposed EnvParams in PO_TrafficLightGridEnv
There are a bunch of env_params in PO_TrafficLightGridEnv that are sort of hidden in the init and not easily visible to users. We should move them out into a config similar to how bottleneck_env works

kjang96:
Closes #145. Added checks to the `__init__` for both classes in the `green_wave_env` file. Edited the experiment runner code in `green_wave.py` to express these changes

### DIFF
--- a/examples/rllab/green_wave.py
+++ b/examples/rllab/green_wave.py
@@ -108,7 +108,7 @@ def run_task(*_):
         "target_velocity": 50,
         "switch_time": 3.0,
         "num_observed": 2,
-        "discrete": False, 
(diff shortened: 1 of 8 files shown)


eugenevinitsky:
Did you add the required params to the rllib examples as well?

eugenevinitsky:
@kjang96 I think all of the grid experiments everywhere will fail unless you add these params to their additional_env_params now

PR #165, opened 2018-09-21
 New sumo - do not merge

eugenevinitsky:


flow/controllers/car_following_models.py
```diff
@@ -381,11 +381,22 @@ def get_accel(self, env):
...
-        # junctions. Setting them to 0 causes vehicles to not move; therefore,
-        # we maintain these negative headways to let sumo control the dynamics
-        # as it sees fit at these points.
+        edge = env.vehicles.get_edge(self.veh_id)
+        lead_edge = env.vehicles.get_edge(lead_id)
+
+        # This covers situations where both vehicles are in a junction from
+        # either direction (e.g. in the figure eight). The vehicle that is
+        # deeper into the junction has its accelerations defined by sumo. This
+        # needed is due to modifications in sumo from 1d4338ab80 to b6f37dd
```
eugenevinitsky: commit number in this line does not seem correct? should be 016c09d306

eugenevinitsky: Also, why would you want only the lead vehicle to have its acceleration defined by sumo? Why not both?

AboudyKreidieh: As in have the acceleration of vehicles in a junction be defined by sumo? I can try that

eugenevinitsky: I mean, you're already doing it for the lead vehicle?

AboudyKreidieh: I actually like your idea better, the solution is more cleaner and more elegant, and the behavior is the same. Will push changes now

eugenevinitsky:
Current errors that must be resolved before merge
- [] bottleneck rllib experiments do not run
- [] the sumo grid example has conflicting directions for the traffic lights
- [] cars go black in the bay bridge as they pass the toll light

eugenevinitsky:
Grid: traffic phases must now be GrGr and yryr instead of the 12 digit strings from before

AboudyKreidieh:
there is a one step lag for adding/removing subscriptions and reading data from the subscriptions. Modifications to the vehicles class will resolve the reason why the bottleneck experiment was not running

AboudyKreidieh:
Update: I modified the vehicles class to immediately return subscription results from new vehicles. This resolved the "index out of range" problem in the bay bridge.

eugenevinitsky:
Bottleneck capacity curve:
image

AboudyKreidieh:
merged relevant changes to `new_sumo_pr`, so closing

PR #167, opened 2018-09-21
fix ray and ppo imports

kev-chien:
fixed ray and ppo imports in tutorial 2 and a stress_test_rl.py


kev-chien:
accidentally changed python version in one line will redo PR

PR #179, opened 2018-09-29
Initial implementation pass of Vehicle subclass for generalized Keras…

crystalzyan:
@fywu85 Initial implementation of Vehicle subclass for generalized Keras-style kernel API, many methods have been skipped. Didn't test it but compiling goes through.


eugenevinitsky:
Oh! This is a neat idea. Obviously, some methods are missing but I'm down to rework things in this manner.

AboudyKreidieh:
sort of stale (there is another version in progress). will close to minimize PR clutter

PR #180, opened 2018-09-30
CORL final version for flow 0.3.0

eugenevinitsky:
- A commit for the release of Flow 0.3.0 to freeze the code for CORL 2018


flow/benchmarks/baselines/merge012.py
```diff
@@ -64,8 +64,8 @@ class needed to run simulations
                departLane="free", departSpeed=7.5)
 
     sumo_params = SumoParams(
-        restart_instance=False,
-        sim_step=0.2,  # time step decreased to prevent occasional crashes
+        restart_instance=True,
+        sim_step=0.5,  # time step decreased to prevent occasional crashes
```
AboudyKreidieh: is this a mistake or something we want to do?

flow/config_default.py
```diff
@@ -16,4 +16,4 @@
 
 AWS_ACCESS_SECRET = os.environ.get("AWS_ACCESS_SECRET", None)
 
-AWS_S3_PATH = "s3://bucket_name"
+AWS_S3_PATH = "s3://public.flow.results/corl_exps/exps_final/ppo/grid_0"
```
AboudyKreidieh: change back to "bucket_name"

flow/visualize/visualizer_rllib.py
```diff
@@ -110,6 +110,7 @@
     module = __import__("flow.envs", fromlist=[flow_params["env_name"]])
     env_class = getattr(module, flow_params["env_name"])
     env_params = flow_params['env']
+    env_params.evaluate = True
```
AboudyKreidieh: should we make this an input argument?

eugenevinitsky: Yeah, we should. It should probably default to false

eugenevinitsky: Otherwise, people will be confused why their trained reward is off

PR #187, opened 2018-10-04
Add experimental renderer based on pyglet. Do not merge.

fywu85:
Only works in sugiyama, loop merge, figure eight, and kathy roundabout for now. Large envs such as bottleneck and baybridge will exceed the renderer max size and crash the program.


fywu85:
Close favoring PR #261

PR #188, opened 2018-10-05
Save rllib algorithm and accessing param in visualizer_rllib

### LINKED ISSUE #173: Save algorithm name in RLlib to ease replay
When we replay an rllib algorithm, we add a --run ALG_NAME flag. We could just save the algorithm name in json for ease of use later and remove the flag

kev-chien:
Closes #173 
Key changes:
- `visualizer_rllib.py` now attempts to look for `flow_params["run"]` (parsed from params.json in result of an rllib experiment) which stores the algorithm used to run the experiment (e.g. "PPO")
- if `flow_params["run"]` is not found, passing in --run ALG when running visualizer_rllib is still an option. This ensures that results from running previous rllib experiments that didn't store the "run" param can still be visualized. 
- Proposed usage: `examples/rllib/stabilizing_the_ring.py` now has `run` as a param (storing algorithm name "PPO"), and changes were made so that one did not need to import the ppo agent and also type the name of the agent "PPO", but only needed to define the agent name "PPO" once in flow_params and then the rest of the example imports the appropriate agent programmatically and stores it.


examples/rllib/stabilizing_the_ring.py
```diff
@@ -86,12 +86,19 @@
     # parameters specifying the positioning of vehicles upon initialization/
     # reset (see flow.core.params.InitialConfig)
     initial=InitialConfig(),
+
```
eugenevinitsky: Hmmm, I'm not sure we want to add this as a flow_param; this adds a param that's rllib specific. Could we keep this same idea, but move it out of flow_params?

eugenevinitsky:
LGTM minus the comment about moving run out of flow_params. Nice work!

eugenevinitsky:
Also, this is failing pep8 tests. You can see where it failed by examining the travis build.

kev-chien:
Currently suggests new usage of storing algorithm name in `rllib/stabilizing_the_ring.py` example.
If it looks good, I can modify the other examples and tutorials

eugenevinitsky:
Yeah, I think this looks good! We could add this to everything in the rllib folder and benchmarks/rllib!

eugenevinitsky:
@kev-chien it's not passing Pep8 requirements. You can see the error by clicking on details on the travis

flow/visualize/visualizer_rllib.py
```diff
@@ -88,7 +90,27 @@
         params=flow_params, version=0, render=False)
     register_env(env_name, create_env)
 
-    agent_cls = get_agent_class(args.run)
+    # Determine agent and checkpoint
+    config_run = config['env_config']['run'] if 'run' in config['env_config'] \
```
eugenevinitsky: you can  write this as config_run = config['env_config'].get('run', None)

PR #189, opened 2018-10-05
Pydocstyle

### LINKED ISSUE #42: documentation standards
In order to ensure that our documentation doesn't degrade with time, we should consider adding tests to ensure compliance with standards such as PEP-257.

One really good tool is `pydocstyle`. To install, run:
```
pip install pydocstyle
```

Then to use while testing for numpy-style documentation (which is what we are following), run:
```
pydocstyle --convention=numpy
```

Our documentation currently needs a lot of work to get it complying with these standards, but once it is up to date, we should add it to the travis tests, as we did with `flake8` for pep8

AboudyKreidieh:
- fixed remaining pydocstyle incompatibilities
- added a test to travis to check for proper documentation
- resolves #42


AboudyKreidieh:
this is very stale, and will be resolved in an upcoming release

PR #202, opened 2018-10-20
Coverage

AboudyKreidieh:
- added coverage badge to track how much of our codebase is hit by unit tests
- we are only at 75%... time to write tests!


AboudyKreidieh:
May God have mercy on us...

PR #208, opened 2018-10-24
removed 'state' from compute_reward

### LINKED ISSUE #191: Remove state as an argument from compute reward
The state doesn't need to be used to produce a reward — benchmark envs shouldn't use the state to compute the reward because a benchmark solution might change the observation function.

AboudyKreidieh:
resolves #191


eugenevinitsky:
I'm not sure why this merge is blocked? It seems like I've approved it...

AboudyKreidieh:
umm, it seems to be merge blocked by nish. I'm just gonna do the evil thing..

PR #222, opened 2018-11-02
All examples are in working condition

kjang96:
All examples are in working condition, added bay_bridge and bay_bridge_toll to test_examples.py.

Deleted obsolete examples.


examples/rllab/stabilizing_the_ring.py
```diff
@@ -18,7 +18,7 @@
 
 def run_task(*_):
     """Implement the run_task method needed to run experiments with rllab."""
-    sumo_params = SumoParams(sim_step=0.1, render=False, seed=0)
+    sumo_params = SumoParams(sim_step=0.1, render=True, seed=0)
```
eugenevinitsky: render should remain false here; we don't want RL experiments to be visualized.

eugenevinitsky:
Duplicated by #223 so I'm closing (I can't push changes to your fork)

PR #238, opened 2018-11-14
Update README.md

eugenevinitsky:
Updated author list on readme


eugenevinitsky:
Yeah I'm not a fan either but I guess we can add him or whatever

PR #245, opened 2018-11-16
Tests base env

AboudyKreidieh:
changes:
- removed a hack from base_env
- added tests to the abstraction in base_env
- added tests to the coloring aspect of base_env

### DIFF
--- a/tests/fast_tests/test_environment_base_class.py
+++ b/tests/fast_tests/test_environment_base_class.py
@@ -16,6 +16,11 @@ import numpy as np
 
 os.environ["TEST_FLAG"] = "True"
 
+# colors for vehicles
+WHITE = (255, 255, 255, 255)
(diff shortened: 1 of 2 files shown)


tests/fast_tests/test_environment_base_class.py
```diff
@@ -414,5 +416,112 @@ def test_it_works(self):
...
+
+        # update the colors of all vehicles
+        env.update_vehicle_colors()
+        env.traci_connection.simulationStep()
+
+        # check the colors of all vehicles
+        for veh_id in env.vehicles.get_ids():
+            if veh_id == "human_0":
+                self.assertEqual(env.traci_connection.vehicle.getColor(veh_id),
+                                 (0, 255, 255, 255))
```
eugenevinitsky: Can we convert these magic number colors to names?

PR #253, opened 2018-11-18
Kernel abstractions

AboudyKreidieh:
This PR contains all abstractions to the kernel class, and is meant to give a sense of the overall architecture of what is to come. All methods current raise `NotImplmenetedError`, and the bulk of the changes are actually documentation. The person reading this is simply meant to read the docstrings and comment on the chose of design. In its current form, the kernel consists of four "sub-kernels" covering simulations, scenarios, traffic lights, and kernels. More details can be found within the PR itself.


mtgibson2014:
@mtgibson2014  is doing this.

flow/core/kernel/kernel.py
```diff
@@ -0,0 +1,94 @@
...
+    def pass_api(self, kernel_api):
+        """Pass the kernel API to all kernel subclasses."""
+        self.kernel_api = kernel_api
+        self.simulation.pass_api(kernel_api)
+        self.scenario.pass_api(kernel_api)
+        self.vehicle.pass_api(kernel_api)
+        self.traffic_light.pass_api(kernel_api)
+
+    def update(self, reset):
+        """Update the kernel subclasses with after a simulation step.
```
mtgibson2014: Should this be: "Update the kernel subclasses after a simulation step"

AboudyKreidieh: done

flow/core/kernel/kernel.py
```diff
@@ -0,0 +1,94 @@
...
+    >>> veh_id = "..."  # some vehicle ID
+    >>> speed = k.vehicle.get_speed(veh_id)
+
+    In addition, these subclasses support sending commands to the simulator via
+    its API. For example, in order to assign a specific vehicle a target
+    acceleration, type:
+
+    >>> k = Kernel(simulator="...")  # a kernel for some simulator type
+    >>> veh_id = "..."  # some vehicle ID
+    >>> k.vehicle.apply_acceleration(veh_id)
```
eugenevinitsky: apply_acceleration is moving to the vehicle class? Which class subclasses GymEnv?

flow/core/kernel/kernel.py
```diff
@@ -0,0 +1,94 @@
...
+            self.simulation = None
+            self.scenario = None
+            self.vehicle = None
+            self.traffic_light = None
+        else:
+            raise ValueError('Simulator type "{}" is not valid.'.
+                             format(simulator))
+
+    def pass_api(self, kernel_api):
+        """Pass the kernel API to all kernel subclasses."""
```
eugenevinitsky: Shouldn't this be called in the init? Otherwise where do they get access to this?

flow/core/kernel/kernel.py
eugenevinitsky: Also, not sure if I agree with the variable name k for the kernel. While it's abundantly clear to us what it means, it may add additional confusion for new developers.

flow/core/kernel/scenario/base.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def next_edge(self, edge, lane):
+        """Return the next edge/lane pair from the given edge/lane.
+
+        These edges may also be internal links (junctions). Returns an empty
+        list if there are no edge/lane pairs in front.
+        """
+        raise NotImplementedError
+
+    def prev_edge(self, edge, lane):
```
eugenevinitsky: Since we're redoing this right now, we may want to discuss what the appropriate thing to do is when two edges converge at a single edge

flow/core/kernel/simulation/base.py
```diff
@@ -0,0 +1,85 @@
+"""Script containing the base simulation kernel class."""
+
```
eugenevinitsky: Repeat; does this become the gym Env?

eugenevinitsky:
Looks good but we should discuss the raised issues.

flow/core/kernel/kernel.py
AboudyKreidieh: So the gym env will still be in base env, with all the relevant calls available from there. What this is meant to do is when you pass a command to the simulator, the command isn't sent to a method within the gym class, but rather to a method within the correct simulation kernel. That way `apply_acceleration` can work for any simulator without us having to modify the envs.

AboudyKreidieh: In terms of k, we'll have documentation on this, but if you'd rather we call it something else, e.g. `self.kernel`, I'd be down too. It just makes calling commands longer

flow/core/kernel/kernel.py
AboudyKreidieh: Init actually is going to consist of three stages:
- create the scenario files
- start the simulation and generate the kernel api
- pass the kernel api to the kernel and all kernel subclasses

flow/core/kernel/scenario/base.py
AboudyKreidieh: sure! this is just architecturally what things will look like, but we can have this discussion when we start adding actual code

eugenevinitsky:
Okay, this looks fine to me. I'm approving it and you can merge it when you're ready.

PR #254, opened 2018-11-18
Vehicle params

AboudyKreidieh:
In this PR, the sumo speed and lane change modes are moved from the `add` method of the vehicles class and into the `SumoCarFollowingParams` and `SumoLaneChangeParams` objects, respectively. This is meant to move all sumo-specific vehicle parameters to a few set of objects, and keep it out of the `Vehicles` class as a whole.

Example usage: 

Before:

```
vehicles.add(
    veh_id="test",
    sumo_car_following_params=SumoCarFollowingParams(),
    sumo_lc_params=SumoLaneChangeParams(),
    speed_mode=9,
    lane_change_mode=0
)
```

After:

```
vehicles.add(
    veh_id="test",
    sumo_car_following_params=SumoCarFollowingParams(
        speed_mode=9,
    ),
    sumo_lc_params=SumoLaneChangeParams(
        lane_change_mode=0,
    ),
)
```


eugenevinitsky:
@AboudyKreidieh this is a really good call; will review this as soon as the build is passing.

examples/rllib/multiagent_exps/multiagent_figure_eight.py
```diff
@@ -42,13 +43,17 @@
         'noise': 0.2
     }),
     routing_controller=(ContinuousRouter, {}),
-    speed_mode='no_collide',
+    sumo_car_following_params=SumoCarFollowingParams(
```
eugenevinitsky: nit: if we're using double quotes elsewhere we should use them here

examples/rllib/multiagent_exps/multiagent_figure_eight.py
```diff
@@ -42,13 +43,17 @@
...
+        speed_mode='no_collide',
```
eugenevinitsky: as above

examples/rllib/multiagent_exps/multiagent_figure_eight.py
```diff
@@ -42,13 +43,17 @@
...
+    ),
     num_vehicles=13)
 vehicles.add(
     veh_id='rl',
     acceleration_controller=(RLController, {}),
     routing_controller=(ContinuousRouter, {}),
-    speed_mode='no_collide',
+    sumo_car_following_params=SumoCarFollowingParams(
```
eugenevinitsky: as above

flow/core/params.py
```diff
@@ -334,6 +344,22 @@ def __init__(
...
         Attributes
         ----------
+        speed_mode : str or int, optional
+            may be one of the following:
+
+             * "right_of_way" (default): respect safe speed, right of way and
+               brake hard at red lights if needed. DOES NOT respect
+               max accel and decel which enables emergency stopping.
+               Necessary to prevent custom models from crashing
+             * "no_collide": Human and RL cars are preventing from reaching
```
eugenevinitsky: Add a note: this may lead to collisions in complex networks

examples/rllib/multiagent_exps/multiagent_figure_eight.py
AboudyKreidieh: If we want to transition to single quotes, I think we should fix lines as we come across them. Or alternatively we can make one PR where we modify all of them. I'm down for both

AboudyKreidieh:
@eugenevinitsky in terms of your comment on the speed mode, I created an issue about in, and will resolve in another PR

PR #255, opened 2018-11-18
Vehicle params 2 (do not merge)

AboudyKreidieh:
Prerequisite: #254 

In this PR, I moved `initial_speed` from the `Vehicles` class into `gen_start_positions`. This is mostly a cleanup, and also because it may cause problems with other simulators (unclear but it seems like a better idea to have it be part of the scenario).


examples/rllab/velocity_bottleneck.py
```diff
@@ -35,10 +35,15 @@
...
     veh_id="human",
-    speed_mode=9,
     lane_change_controller=(SumoLaneChangeController, {}),
     routing_controller=(ContinuousRouter, {}),
-    lane_change_mode=0,  # 1621,#0b100000101,
+    sumo_car_following_params=SumoCarFollowingParams(
+        speed_mode=9,
+    ),
+    sumo_lc_params=SumoLaneChangeParams(
+        lane_change_mode=0,  # 1621,#0b100000101,
```
mtgibson2014: What is this comment for?

examples/sumo/bay_bridge.py
```diff
@@ -43,12 +43,16 @@ def bay_bridge_example(render=None,
     if render is not None:
         sumo_params.render = render
 
-    sumo_car_following_params = SumoCarFollowingParams(speedDev=0.2)
+    sumo_car_following_params = SumoCarFollowingParams(
+        speedDev=0.2,
+        speed_mode="all_checks",
+    )
     sumo_lc_params = SumoLaneChangeParams(
```
mtgibson2014: I noticed that sometimes you perform this line of code in the "vechicle.add()" below - instead of performing this line separately as you do right here. Not sure if you wanted to be consistent with this or if it doesn't matter.

eugenevinitsky:
I'm not sure I agree with this; I don't know if users will likely assume that they have to specify car starting speeds in scenarios instead of in the vehicles.

AboudyKreidieh:
Hmm, that's a good point. Looking at the changes to the kernel I don't think I see any significant value. Will try to modify the upcoming PRs without these changes and get back to you on this.

AboudyKreidieh:
worked things out without this. closing it, but will create the branch around for a while

PR #256, opened 2018-11-18
Base env cleanup

AboudyKreidieh:
changes:
- removed some traci calls from base_env
- removed some things in base_env that are never used or called


flow/envs/base_env.py
```diff
@@ -349,8 +314,20 @@ def setup_initial_state(self):
         self.vehicles.update(vehicle_obs, id_lists, self)
         self.traffic_lights.update(tls_obs)
 
-        # store the network observations in the vehicles class
-        self.vehicles.update(vehicle_obs, id_lists, self)
+        # check to make sure all vehicles have been spawned
+        if len(self.initial_ids) < self.vehicles.num_vehicles:
+            logging.error("Not enough vehicles have spawned! Bad start?")
+            exit()
```
42B: Shouldn't this use `sys.exit` and return a nonzero value?

eugenevinitsky: This is correct, thanks for the catch!

42B: @eugenevinitsky 

No problem!

flow/envs/base_env.py
```diff
@@ -323,12 +323,6 @@ def setup_initial_state(self):
             value = sparse state information (only what is needed to add a
             vehicle in a sumo network with traci)
         """
-        # check to make sure all vehicles have been spawned
-        num_spawned_veh = self.traci_connection.simulation.getDepartedNumber()
```
eugenevinitsky: why is this a bad check to have in place?

flow/envs/multiagent_env.py
```diff
@@ -261,7 +254,7 @@ def reset(self):
                     route_id,
                     typeID=str(type_id),
                     departLane=str(lane_index),
```
eugenevinitsky: Thank you for copying this!

flow/envs/base_env.py
AboudyKreidieh: Don’t worry, it was just shifted down. The check is still there.

PR #267, opened 2018-11-20
Deep copy vehicles variable in create_env, which will allow for clean…

kjang96:
Deep copy vehicles variable in create_env, which will allow for clean resets. Currently, old parameters are persisting beyond their intended iteration.

### DIFF
--- a/flow/utils/registry.py
+++ b/flow/utils/registry.py
@@ -67,6 +67,9 @@ def make_create_env(params, version=0, render=None):
     traffic_lights = params.get("tls", TrafficLights())
 
     def create_env(*_):
+        vehicles = deepcopy(params['veh'])
+        sumo_params = deepcopy(params['sumo'])
+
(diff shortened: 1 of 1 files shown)


flow/utils/registry.py
```diff
@@ -62,11 +62,13 @@ def make_create_env(params, version=0, render=None):
 
     env_params = params['env']
     net_params = params['net']
-    vehicles = params['veh']
     initial_config = params.get('initial', InitialConfig())
     traffic_lights = params.get("tls", TrafficLights())
 
     def create_env(*_):
+        sumo_params = deepcopy(params['sumo'])
+        vehicles = deepcopy(params['vehicles'])
```
AboudyKreidieh: rename 'vehicles' -> 'veh'

kjang96: Done

PR #270, opened 2018-11-20
Multiagent pr 2

eugenevinitsky:
Separate PR for the multi-agent upgrades. @AboudyKreidieh I felt for you.

### DIFF
--- a/examples/rllib/multiagent_exps/multiagent_figure_eight.py
+++ b/examples/rllib/multiagent_exps/multiagent_figure_eight.py
@@ -1,0 +1,158 @@
+"""Example of a multi-agent environment containing a figure eight with
+one autonomous vehicle and an adversary that is allowed to perturb
+the accelerations of figure eight."""
+
+# WARNING: Expected total reward is zero as adversary reward is
+# the negative of the AV reward
+
+from copy import deepcopy
+import json
+import os
+
+import ray
+import ray.rllib.agents.ppo as ppo
+from ray.rllib.agents.ppo.ppo_policy_graph import PPOPolicyGraph
+from ray import tune
+from ray.tune.registry import register_env
+from ray.tune import run_experiments
+
+from flow.controllers import ContinuousRouter
+from flow.controllers import IDMController
+from flow.controllers import RLController
+from flow.core.params import EnvParams
+from flow.core.params import InitialConfig
+from flow.core.params import NetParams
+from flow.core.params import SumoParams
+from flow.core.vehicles import Vehicles
+from flow.scenarios.figure_eight import ADDITIONAL_NET_PARAMS
+from flow.utils.registry import make_create_env
+from flow.utils.rllib import FlowParamsEncoder
+
+# time horizon of a single rollout
+HORIZON = 1500
+# number of rollouts per training iteration
+N_ROLLOUTS = 14
+# number of parallel workers
+N_CPUS = 14
+os.environ['MULTIAGENT'] = 'True'
+
+# We place one autonomous vehicle and 13 human-driven vehicles in the network
+vehicles = Vehicles()
+vehicles.add(
+    veh_id='human',
+    acceleration_controller=(IDMController, {
+        'noise': 0.2
+    }),
+    routing_controller=(ContinuousRouter, {}),
+    speed_mode='no_collide',
+    num_vehicles=13)
+vehicles.add(
+    veh_id='rl',
+    acceleration_controller=(RLController, {}),
+    routing_controller=(ContinuousRouter, {}),
+    speed_mode='no_collide',
+    num_vehicles=1)
+
+flow_params = dict(
+    # name of the experiment
+    exp_tag='multiagent_figure_eight',
+
+    # name of the flow environment the experiment is running on
+    env_name='MultiAgentAccelEnv',
+
+    # name of the scenario class the experiment is running on
+    scenario='Figure8Scenario',
+
+    # sumo-related parameters (see flow.core.params.SumoParams)
+    sumo=SumoParams(
+        sim_step=0.1,
+        render=False,
+    ),
+
+    # environment related parameters (see flow.core.params.EnvParams)
+    env=EnvParams(
+        horizon=HORIZON,
+        additional_params={
+            'target_velocity': 20,
+            'max_accel': 3,
+            'max_decel': 3,
+            'perturb_weight': 0.03
+        },
+    ),
+
+    # network-related parameters (see flow.core.params.NetParams and the
+    # scenario's documentation or ADDITIONAL_NET_PARAMS component)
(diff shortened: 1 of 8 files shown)


eugenevinitsky:
Don't merge; I just wanted to check that the tests will pass. Will be ready for merge tonight after additional tests are added.

flow/benchmarks/figureeight0.py
```diff
@@ -66,7 +67,7 @@
     # scenario's documentation or ADDITIONAL_NET_PARAMS component)
     net=NetParams(
         no_internal_links=False,
-        additional_params=ADDITIONAL_NET_PARAMS,
+        additional_params=deepcopy(ADDITIONAL_NET_PARAMS),
```
AboudyKreidieh: nit: does `ADDITIONAL_NET_PARAMS.copy()` not work here? u used it once above, and from my experience that works fine too

flow/core/vehicles.py
```diff
@@ -24,7 +24,6 @@
 
 class Vehicles:
     """Base vehicle class.
-
```
AboudyKreidieh: all the deleted spaces in this file should not have been deleted as per documentation standards

flow/core/vehicles.py
```diff
@@ -971,44 +926,83 @@ def set_lane_headways(self, veh_id, lane_headways):
...
+        list<float>
+        """
+        if isinstance(veh_id, (list, np.ndarray)):
+            return [self.get_lane_leaders_speed(vehID, error)
+                    for vehID in veh_id]
+        return self.__vehicles.get(veh_id, {}).get("leaders_speed", error)
+
+    def set_lane_leaders_speed(self, veh_id, lead_speed):
+        """Set the speed of the leaders of the specified vehicles"""
+        self.__vehicles[veh_id]["leaders_speed"] = lead_speed
```
AboudyKreidieh: i don't think there's any point of storing these things like this. u can simply call `self.get_speed(self.get_lane_leaders(veh_id))` in this method

flow/core/vehicles.py
```diff
@@ -971,44 +926,83 @@ def set_lane_headways(self, veh_id, lane_headways):
...
+        error : any, optional
+            value that is returned if the vehicle is not found
+        Returns
+        -------
+        list<float>
+        """
+        if isinstance(veh_id, (list, np.ndarray)):
+            return [self.get_lane_leaders_speed(vehID, error)
+                    for vehID in veh_id]
+        return self.__vehicles.get(veh_id, {}).get("leaders_speed", error)
```
AboudyKreidieh: i don't think there's any point of storing these things like this. u can simply call `self.get_speed(self.get_lane_leaders(veh_id))` in this method

flow/core/vehicles.py
```diff
@@ -971,44 +926,83 @@ def set_lane_headways(self, veh_id, lane_headways):
...
+        error : any, optional
+            value that is returned if the vehicle is not found
+        Returns
+        -------
+        list<float>
+        """
+        if isinstance(veh_id, (list, np.ndarray)):
+            return [self.get_lane_followers_speed(vehID, error)
+                    for vehID in veh_id]
+        return self.__vehicles.get(veh_id, {}).get("followers_speed", error)
```
AboudyKreidieh: same as above

flow/scenarios/multi_loop.py
```diff
@@ -0,0 +1,309 @@
...
+        r = length / (2 * pi)
+        ring_spacing = 4 * r
+        num_rows = num_cols = int(ceil(sqrt(ring_num)))
+
+        nodes = []
+        i = 0
+        for j in range(num_rows):
+            for k in range(num_cols):
+                nodes += [{
+                    "id": "bottom_{}".format(i),
```
AboudyKreidieh: many double quotes. want to fix now or in a later commit?

flow/envs/multiagent_env.py
```diff
@@ -0,0 +1,345 @@
...
+        observation: dict of numpy ndarrays
+            agent's observation of the current environment
+        reward: dict of floats
+            amount of reward associated with the previous state/action pair
+        done: dict of bools
+            indicates whether the episode has ended
+        info: dict
+            contains other diagnostic information from the previous action
+        """
+        for _ in range(self.env_params.sims_per_step):
```
AboudyKreidieh: is this just duplicated from changes in `base_env.py`? if so, should we delete one?

flow/scenarios/multi_loop.py
eugenevinitsky: Later commit; I'll do one 100% only double quotes PR so that you don't even have to read it to approve.

flow/envs/multiagent_env.py
eugenevinitsky: So; it's not just duplicated, there is a slightly different part at the bottom. It's mostly duplicated and I can't figure out how to not reuse this code.

flow/core/vehicles.py
eugenevinitsky: It's just a convenience method to make it easier for users so they don't have to chain these two methods together

flow/core/vehicles.py
eugenevinitsky: Ty, will fix.

eugenevinitsky:
image

HAHAHA

eugenevinitsky:
I'll also add some descriptions of this to the documentation first

flow/core/vehicles.py
eugenevinitsky: @AboudyKreidieh we should formalize these standards in a customized style checker.

eugenevinitsky:
@AboudyKreidieh we should have this discussion before I merge:
MultiEnv is for when you have multiple agents, everything else just subclasses Env. I am adding examples of this to the docs. There doesn't appear to be a cleaner solution at the moment.

flow/core/vehicles.py
AboudyKreidieh: yup, this is part of #189 , but it's been low priority for a while

flow/core/vehicles.py
AboudyKreidieh: right, but from within this method we can chain them, and not add all the extra code to the vehicles class

flow/scenarios/multi_loop.py
AboudyKreidieh: sounds good!

flow/envs/multiagent_env.py
AboudyKreidieh: kk, we still will need a tutorial or something that describes when to use what

flow/core/vehicles.py
eugenevinitsky: Got it; resolved.

PR #276, opened 2018-11-25
Update visualizing.rst

eugenevinitsky:
This file is out of date and is now updated.


eugenevinitsky:
This decreases the coverage slightly but only because coveralls seems to be including our docs....

PR #278, opened 2018-11-25
Sumo web3d upgrade

eugenevinitsky:
- added ability to use sumo_web3d to visualize
- added documentation for how to use it


flow/visualize/visualizer_rllib.py
```diff
@@ -43,7 +43,7 @@ def visualizer_rllib(args):
     config = get_rllib_config(result_dir)
 
     # Run on only one cpu for rendering purposes
-    config['num_workers'] = 1
+    config['num_workers'] = 0
```
AboudyKreidieh: why did this number change?

PR #281, opened 2018-11-27
Kernel simulation

AboudyKreidieh:
Prerequisite: #253 

changes:
- added traci vehicle simulation kernel
- replaced `start_sumo` with kernel call
- replaced simulation-based traci calls with kernel calls


flow/envs/base_env.py
```diff
@@ -136,10 +128,18 @@ def __init__(self, env_params, sumo_params, scenario):
...
 
-        # contains the subprocess.Popen instance used to start traci
-        self.sumo_proc = None
+        # create the Flow kernel
+        self.k = Kernel(simulator="traci", sim_params=sumo_params)
+
+        # initialize the simulation using the simulation kernel. This will use
+        # the scenario kernel as an input in order to determine what network
+        # needs to be simulated.
+        self.traci_connection = self.k.simulation.start_simulation(
```
eugenevinitsky: er, probably shouldn't be named traci_connection

flow/envs/base_env.py
```diff
@@ -199,113 +199,11 @@ def restart_sumo(self, sumo_params, render=None):
             ensure_dir(sumo_params.emission_path)
             self.sumo_params.emission_path = sumo_params.emission_path
 
-        self.start_sumo()
+        self.traci_connection = self.k.simulation.start_simulation(
```
eugenevinitsky: as above

flow/envs/base_env.py
AboudyKreidieh: totally agree, but traci_connection is used throughout the codebase, so this is just a temporary fix as we weed them out in later PRs.

PR #285, opened 2018-11-28
Redis error

eugenevinitsky:
Redis incompatibilities in rllib are addressed in the travis.yml but not in environment.yml. This PR addresses this.


eugenevinitsky:
#283 is a prerequisite

PR #292, opened 2018-11-30
solve naming conflict for new sumo

yunerzxy:
examples/sumo/bottleneck.py -> examples/sumo/bottlenecks.py


eugenevinitsky:
@yunerzxy what error is this solving? Can you create an issue for it so we can track it?

PR #295, opened 2018-12-01
Minor bug fixes

fywu85:
1. Rename `bottleneck.py` to `bottlenecks.py` in `examples/sumo/` to avoid name conflict with pandas.
2. Properly reset `lanes` in the reset method of `WaveAttenuationEnv`.


fywu85:
Just noticed this. The first fix is essentially #292.

PR #309, opened 2018-12-13
some trivial edit.

CYBruce:
Actually, the edge0 is not a node (from my understanding). So, I think this sentence may confuse some new reader like me.


AboudyKreidieh:
yup! LGTM

PR #310, opened 2018-12-13
Kernel traffic lights

AboudyKreidieh:
Prerequisite: #281 

changes:
- moved the `TrafficLights` class to params.py (it is no longer a meaningful object, but rather a parameter)
- created the traci traffic lights kernel subclass
- replaced calls to the traffic light class will calls to the kernel subclass


examples/rllab/green_wave.py
```diff
@@ -6,7 +6,7 @@
 from rllab.envs.gym_env import GymEnv
 
 from flow.core.vehicles import Vehicles
-from flow.core.traffic_lights import TrafficLights
+from flow.core.params import TrafficLights
```
eugenevinitsky: We should renamed this TrafficLightsParams then

flow/core/kernel/kernel.py
```diff
@@ -53,10 +60,10 @@ def __init__(self, simulator):
...
 
         if simulator == "traci":
-            self.simulation = None
-            self.scenario = None
-            self.vehicle = None
-            self.traffic_light = None
+            self.simulation = TraCISimulation(self)
+            self.scenario = KernelScenario(self)
+            self.vehicle = KernelVehicle(self, sim_params)
+            self.traffic_light = TraCITrafficLight(self)
```
eugenevinitsky: Add FIXME so we don't forget to make this non-traci as well

flow/core/kernel/kernel.py
```diff
@@ -83,9 +90,9 @@ def update(self, reset):
             specifies whether the simulator was reset in the last simulation
             step
         """
-        self.scenario.update(reset)
+        # self.scenario.update(reset)
         self.simulation.update(reset)
-        self.vehicle.update(reset)
+        # self.vehicle.update(reset)
```
eugenevinitsky: why?

flow/core/params.py
```diff
@@ -13,6 +13,176 @@
...
 LC_MODES = {"aggressive": 0, "no_lat_collide": 512, "strategic": 1621}
 
+# Traffic light defaults
+PROGRAM_ID = 1
+MAX_GAP = 3.0
+DETECTOR_GAP = 0.6
+SHOW_DETECTORS = True
+
+
+class TrafficLights:
```
eugenevinitsky: Rename please

eugenevinitsky:
This makes sense to me. Will approve once we resolve naming issues?

flow/core/kernel/kernel.py
AboudyKreidieh: don't worry, these are covered in later branched (which will be available to PR after this wave)

flow/core/params.py
AboudyKreidieh: done!

examples/rllab/green_wave.py
AboudyKreidieh: done!

flow/core/kernel/kernel.py
AboudyKreidieh: just a temporary fix until we merge the next wave of PRs

PR #316, opened 2018-12-15
converted gen_custom_start_pos to a staticmethod

AboudyKreidieh:
Once we transition to the kernel, the `gen_custom_start_pos` method needs to be created outside the kernel. This will help cleanly facilitate that.


examples/rllab/green_wave.py
```diff
@@ -33,7 +33,7 @@ def gen_edges(row_num, col_num):
 def get_flow_params(v_enter, vehs_per_hour, col_num, row_num,
                     additional_net_params):
     initial_config = InitialConfig(
-        spacing="uniform", lanes_distribution=float("inf"), shuffle=True)
+        spacing="custom", lanes_distribution=float("inf"), shuffle=True)
```
eugenevinitsky: what's the reasoning for this?

AboudyKreidieh: The grid network actually overrides the uniform starting position method, which causes issues once the kernel scenario gets introduced, so I just replaced `gen_uniform_start_pos` with `gen_custom_start_pos` and modified everything accordingly. Ignoring what might come with the scenario kernel, I still think this is a good idea.

flow/scenarios/grid.py
```diff
@@ -512,21 +512,22 @@ def specify_intersection_edge_starts(self):
             [(":center", 0)]
         return intersection_edgestarts
 
-    def gen_even_start_pos(self, initial_config, num_vehicles, **kwargs):
+    @staticmethod
```
eugenevinitsky: @staticmethod is not needed for class methods, I think the correct @ is @classmethod?

flow/scenarios/highway.py
```diff
@@ -109,25 +109,10 @@ def specify_edge_starts(self):
                       for i in range(self.num_edges)]
         return edgestarts
 
-    def gen_custom_start_pos(self, initial_config, num_vehicles, **kwargs):
+    @staticmethod
```
eugenevinitsky: Should be @classmethod if that is the intent

flow/scenarios/loop_merge.py
```diff
@@ -242,76 +242,67 @@ def specify_internal_edge_starts(self):
 
         return internal_edgestarts
 
-    def gen_custom_start_pos(self, initial_config, num_vehicles, **kwargs):
+    @staticmethod
```
eugenevinitsky: As above.

flow/scenarios/base_scenario.py
```diff
@@ -495,12 +495,15 @@ def gen_random_start_pos(self, initial_config, num_vehicles, **kwargs):
 
         return startpositions, startlanes
 
-    def gen_custom_start_pos(self, initial_config, num_vehicles, **kwargs):
+    @staticmethod
+    def gen_custom_start_pos(cls, initial_config, num_vehicles, **kwargs):
```
eugenevinitsky: Question, do you know how class methods fare if parallelism is used? i.e. do all the threads share the class instance?

PR #317, opened 2018-12-16
removed some traci calls (do not merge)

AboudyKreidieh:
This PR is meant to simplify later kernel modifications by removing some traci calls.

Changes:
- replaced traci traffic light calls with calls to the traffic lights class
- removed color changing in bottleneck and bay bridge envs. This would've only been slowing down simulations anyways


flow/envs/bay_bridge/base.py
```diff
@@ -112,8 +112,6 @@ def ramp_meter_lane_change_control(self):
             if self.vehicles.get_edge(veh_id) == EDGE_AFTER_RAMP_METER:
                 lane_change_mode = self.cars_before_ramp[veh_id][
                     "lane_change_mode"]
-                color = self.cars_before_ramp[veh_id]["color"]
```
eugenevinitsky: Hmm? This has a function; it lets you visually identify if the toll booth is on.

AboudyKreidieh: okay, let's hold off on this. I can add an API call for changing colors and it'll be fine

PR #320, opened 2018-12-18
Scenario compatibility

AboudyKreidieh:
Prerequisite: #319 

changes:
- modified definition of node, edge, connections, and types to use integers and floats instead of strings
- modified the definition of shapes in edges to include lists of (x,y) tuples instead of strings
- modified the base scenario class to convert the integers, floats, and shapes to be compatible with sumo.

### DIFF
--- a/flow/scenarios/minicity.py
+++ b/flow/scenarios/minicity.py
@@ -420,8 +420,8 @@ class MiniCityScenario(Scenario):
                     [np.sqrt((edge['shape'][i][0] - edge['shape'][i+1][0])**2 +
                              (edge['shape'][i][1] - edge['shape'][i+1][1])**2)
                      * SCALING for i in range(len(edge['shape'])-1)])
-                edge['shape'] = [(blip*SCALING, blop*SCALING)
-                                 for blip, blop in edge['shape']]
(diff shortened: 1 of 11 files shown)


flow/scenarios/minicity.py
```diff
@@ -415,19 +415,16 @@ def specify_edges(self, net_params):
...
             if 'shape' in edge:
                 edge['length'] = sum(
                     [np.sqrt((edge['shape'][i][0] - edge['shape'][i+1][0])**2 +
                              (edge['shape'][i][1] - edge['shape'][i+1][1])**2)
                      * SCALING for i in range(len(edge['shape'])-1)])
-                edge['length'] = str(edge['length'])
-                edge['shape'] = ' '.join('%.2f,%.2f' % (blip*SCALING,
-                                                        blop*SCALING)
-                                         for blip, blop in edge['shape'])
+                edge['shape'] = [(blip*SCALING, blop*SCALING)
```
eugenevinitsky: uh, can we change this from blip and blop hahaha

AboudyKreidieh: haha, fixed

PR #322, opened 2018-12-20
Enable collision checks at intersections

fywu85:
This will enable collision detection at traffic intersections.


AboudyKreidieh:
Hey @fywu85 what does it mean to enable collision checks? Just curious..

fywu85:
Without this, sumo will not register collisions at intersections. This
means that vehicles can run through each other without being flagged as
collisions.

PR #330, opened 2018-12-22
Params abstraction

AboudyKreidieh:
changes:
- replaced `restart_sumo` -> `restart_simulation`
- replaced `sumo_params` -> `sim_params`
- created a generic `SimParams` that acts as a parent to `SumoParams`, and placed in it all the properties that are not sumo-specific


AboudyKreidieh:
ty!

PR #343, opened 2018-12-26
created objects that are used when passing data from aimsun

AboudyKreidieh:


flow/utils/aimsun/struct.py
```diff
@@ -0,0 +1,222 @@
...
+    segment : int
+        Segment number of the section where the vehicle is located (from 0 to
+        n-1)
+    numberLane : int
+        Lane number in the segment (from 1, the rightmost lane, to N, the
+        leftmost lane)
+    idJunction : int
+        The junction identifier
+    idSectionFrom : int
+        Origin section identifier
```
Yasharzf: maybe add:
```suggestion
        Origin section identifier when the vehicle is in a node        
```

flow/utils/aimsun/struct.py
```diff
@@ -0,0 +1,222 @@
...
+    idLaneFrom : int
+        Origin section´s lane where the vehicle enters the junction from. 1
+        being the rightmost lane and N the leftmost lane, being N the number of
+        lanes in the origin section
+    idSectionTo : int
+        Destination section identifier
```
Yasharzf: maybe add:
```suggestion
        Destination section identifier when the vehicle is in a node 
```

PR #355, opened 2018-12-31
Robustness boost

fywu85:
Attempt to enhance software robustness through introducing the following two features:
- automatic garbage collection in base env
- support recent ray API change

### DIFF
--- a/flow/envs/base_env.py
+++ b/flow/envs/base_env.py
@@ -6,6 +6,7 @@ from gym.spaces import Box
 import logging
 import os
 import sys
+import atexit
 import time
 import traceback
 import numpy as np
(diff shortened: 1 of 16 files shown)


flow/envs/base_env.py
```diff
@@ -879,16 +881,20 @@ def terminate(self):
...
+                "Closing connection to TraCI and stopping simulation.\n"
+                "Note, this may print an error message when it closes."
+            )
+            self.traci_connection.close()
+            self.scenario.close()
+
+            # close pyglet renderer
+            if self.sim_params.render in ['gray', 'dgray', 'rgb', 'drgb']:
+                self.renderer.close()
+        except FileNotFoundError:
```
AboudyKreidieh: when does this happen?

fywu85: If an user manually closes the env, `scenario.close()` will fail since all the net files and cfg files are already gone. Thus the `FileNotFoundError`.

fywu85:
Thx!

flow/envs/base_env.py
AboudyKreidieh: kk, sounds good

PR #358, opened 2019-01-01
moved multiagent environment to a new folder

### LINKED ISSUE #337: Profile performance
There appears to be a performance hit for very easy examples (like the sugiyama example, where RLlib/tensorflow, etc. are not needed), which takes SUMO a while to load. Any ideas where the slowdown might be? This shouldn't be much slower than running the sumo binary directly. It would be good to profile the performance and see if we are loading extra libraries when they aren't needed.

**Note**: This issue may be confounded with #334.

AboudyKreidieh:
resolves #337 

The new multiagent environments enforced a ray dependency, which needs to imported every time an environment is used and takes a while to import.


AboudyKreidieh:
Update: this also ensures that you don't need to install `ray` if u want to run the single agent exps (e.g. if u are using rllab instead)

Issue #365, opened 2019-01-03
Remove cls from custom_start_pos


AboudyKreidieh:
Many things in the custom start pos methods called variables within the scenario class (not kernel), and so I had to replace those calls with `cls.network.*`, with network being the scenario class. This is a bit hacky, so needs to be remedied.

Issue #364, opened 2019-01-03
remove self.scenario from Flow


AboudyKreidieh:
Currently there is a scenario kernel names `self.k.scenario` and the old scenario class `self.scenario`, with the latter serving in a sense as a parameter to the former. However, `self.scenario` is called a lot in Flow, and can lead to confusion. Accordingly, we should remove `self.scenario` from everywhere.

PR #366, opened 2019-01-03
Kernel scenario

AboudyKreidieh:
changes:
- moved the generating starting position methods from the base scenario class to the base scenario kernel.
- custom starting positions within the base scenario class are called by another starting position method within the scenario kernel
- created a sumo scenario kernel, and moved all relevant methods from the base scenario class to the sumo scenario kernel
- the base scenario class now creates separate variables for nodes, edges, types and connections, which are then used by the new scenario kernels
- replaced `self.scenario` with `self.k.scenario` wherever relevant

Some aspects of this PR are hacky and need to be resolved ASAP. These are:
- #365 
- #364


flow/core/kernel/kernel.py
```diff
@@ -1,7 +1,7 @@
 """Script containing the Flow kernel object for interacting with simulators."""
 
 from flow.core.kernel.simulation import TraCISimulation
-from flow.core.kernel.scenario import KernelScenario
+from flow.core.kernel.scenario import TraCIScenario
```
eugenevinitsky: A bit confused by this file structure overall. Will kernel.scenario contain both TraciScenario and AimsunScenario?

flow/core/kernel/scenario/__init__.py
```diff
@@ -1,3 +1,4 @@
 from flow.core.kernel.scenario.base import KernelScenario
+from flow.core.kernel.scenario.traci import TraCIScenario
```
eugenevinitsky: The import path here is different than the import above?

flow/core/kernel/scenario/base.py
```diff
@@ -57,6 +61,12 @@ def __init__(self, master_kernel):
         self.network = None
         self.orig_name = None
         self.name = None
+        self.edgestarts = None
+        self.internal_edgestarts = None
+        self.intersection_edgestarts = None
+        self.internal_edgestarts_dict = None
```
eugenevinitsky: Are we keeping the internal edge_starts object? What function does it serve?

flow/core/kernel/scenario/base.py
```diff
@@ -251,22 +253,97 @@ def gen_even_start_pos(self, initial_config, num_vehicles, **kwargs):
...
+            self._get_start_pos_util(initial_config, num_vehicles)
+
+        # return an empty list of starting positions and lanes if there are no
+        # vehicles to be placed
+        if num_vehicles == 0:
+            return [], []
+
+        increment = available_length / num_vehicles
+
+        # if not all lanes are equal, then we must ensure that vehicles are in
```
eugenevinitsky: This comment doesn't seem correctly worded, I'm not sure what it's trying to get across.

flow/core/kernel/scenario/base.py
```diff
@@ -251,22 +253,97 @@ def gen_even_start_pos(self, initial_config, num_vehicles, **kwargs):
...
+                car_count += 1
+                startpositions.append(pos)
+                startlanes.append(lane)
+
+                if car_count == num_vehicles:
+                    break
+
+            x = (x + increment + VEHICLE_LENGTH + min_gap) % self.length()
+
+        # add a perturbation to each vehicle, while not letting the vehicle
```
eugenevinitsky: Maybe not the right moment, but we should do a check here that this doesn't cause it to overlap onto a nearby vehicle.

flow/core/kernel/scenario/base.py
```diff
@@ -275,22 +352,72 @@ def gen_random_start_pos(self, initial_config, num_vehicles, **kwargs):
...
+
+        for edge in available_edges:
+            available_length -= efs * min([self.num_lanes(edge), lanes_distr])
+
+        # choose random positions for each vehicle
+        init_absolute_pos = \
+            [random.random() * available_length
+             for _ in range(num_vehicles)]
+
+        # sort the positions of vehicles, for simplicity in using
```
eugenevinitsky: Nit: "for simplicity in using", using what? Seems like a word is missing

flow/core/kernel/scenario/base.py
```diff
@@ -251,22 +253,97 @@ def gen_even_start_pos(self, initial_config, num_vehicles, **kwargs):
...
+        if initial_config.perturbation > 0:
+            for i in range(num_vehicles):
+                perturb = np.random.normal(0, initial_config.perturbation)
+                edge, pos = startpositions[i]
+                pos = max(0, min(self.edge_length(edge), pos + perturb))
+                startpositions[i] = (edge, pos)
+
+        return startpositions, startlanes
+
+    def gen_random_start_pos(self, initial_config, num_vehicles):
```
eugenevinitsky: Am I correct in seeing that this method does not check/resolve overlapping vehicle positions?

eugenevinitsky: Note, if not, that such a resolution can be done in N^2 where N is the number of vehicles. Doesn't necessarily have to be done now, but if not it should be added to the issues.

flow/core/kernel/scenario/base.py
```diff
@@ -275,22 +352,72 @@ def gen_random_start_pos(self, initial_config, num_vehicles, **kwargs):
...
+        for i in range(num_vehicles):
+            edge_i = available_edges[edge_indx]
+            pos_i = (init_absolute_pos[i] - decrement) % (
+                    self.edge_length(edge_i) - efs)
+            lane_i = int(((init_absolute_pos[i] - decrement) - pos_i) /
+                         (self.edge_length(edge_i) - efs))
 
-    def gen_custom_start_pos(self, initial_config, num_vehicles, **kwargs):
+            pos_i += efs
+
```
eugenevinitsky: This loop deserves a comment explaining what it does.

flow/core/kernel/scenario/base.py
```diff
@@ -301,17 +428,111 @@ def gen_custom_start_pos(self, initial_config, num_vehicles, **kwargs):
...
+        num_vehicles : int
+            number of vehicles to be placed on the network
+
+        Returns
+        -------
+        x0 : float
+            starting position of the first vehicle, in meters
+        min_gap : float
+            minimum gap between vehicles
+        bunching : float
```
eugenevinitsky: Nit: I think we should adjust this comment, I still don't understand _exactly_ how bunching works.

flow/core/kernel/scenario/traci.py
```diff
@@ -0,0 +1,885 @@
+"""Script containing the TraCI scenario kernel class."""
```
eugenevinitsky: I'm currently assuming this is copied over from other code unchanged, if it is not and needs review, let me know

flow/core/kernel/scenario/traci.py
```diff
@@ -0,0 +1,885 @@
...
+
+        Returns
+        -------
+        edges : dict <dict>
+            Key = name of the edge
+            Elements = length, lanes, speed
+        connection_data : dict < dict < list<tup> > >
+            Key = name of the arriving edge
+                Key = lane index
+                Element = list of edge/lane pairs that a vehicle can traverse
```
eugenevinitsky: Nit: if it's a tuple then this should be written as (edge, lane)

flow/core/kernel/scenario/traci.py
```diff
@@ -0,0 +1,885 @@
...
+
+        # specify the location of the output file
+        netfn = "%s.net.xml" % self.name
+
+        # generate the network file with sumo
+        net_cmd = "netconvert --osm-files {0} --output-file {1}".\
+            format(osm_path, self.cfg_path + netfn)
+
+        # this handles removing all roads in the network that cannot be ridden
+        # by vehicles
```
eugenevinitsky: We should put an issue up about this, presumably at some point we may want to have pedestrians

flow/core/kernel/scenario/traci.py
```diff
@@ -0,0 +1,885 @@
...
+
+        Returns
+        -------
+        edges : dict <dict>
+            Key = name of the edge
+            Elements = length, lanes, speed
+        connection_data : dict < dict < list<tup> > >
+            Key = name of the arriving edge
+                Key = lane index
+                Element = list of edge/lane pairs that a vehicle can traverse
```
eugenevinitsky: Nit: same as before

flow/core/kernel/scenario/traci.py
```diff
@@ -0,0 +1,885 @@
...
+                      'http://sumo.dlr.de/xsd/additional_file.xsd')
+
+        # add the routes to the .add.xml file
+        for (edge, route) in self.rts.items():
+            add.append(E('route', id='route%s' % edge, edges=' '.join(route)))
+
+        # add (optionally) the traffic light properties to the .add.xml file
+        num_traffic_lights = len(list(traffic_lights.get_properties().keys()))
+        if num_traffic_lights > 0:
+            if traffic_lights.baseline:
```
eugenevinitsky: we should rename this; in one case it is called baseline and in the other "actuated_default"; it's confusing

flow/core/params.py
```diff
@@ -1308,7 +1310,7 @@ def _multi_lane_headways_util(self, veh_id, edge_dict, num_edges, env):
         this_pos = self.get_position(veh_id)
         this_edge = self.get_edge(veh_id)
         this_lane = self.get_lane(veh_id)
-        num_lanes = env.scenario.num_lanes(this_edge)
+        num_lanes = env.k.scenario.num_lanes(this_edge)
 
         # set default values for all output values
         headway = [1000] * num_lanes
```
eugenevinitsky: This default value should be moved into the method definition as an optional param; otherwise it's unexpected behavior

flow/core/params.py
```diff
@@ -1421,17 +1423,17 @@ def _prev_edge_followers(self, veh_id, edge_dict, lane, num_edges, env):
         pos = self.get_position(veh_id)
         edge = self.get_edge(veh_id)
 
-        tailway = 1000  # env.scenario.length
+        tailway = 1000
```
eugenevinitsky: Nit: as above

flow/envs/base_env.py
```diff
@@ -103,11 +102,6 @@ def __init__(self, env_params, sim_params, scenario):
         # simulation step size
         self.sim_step = sim_params.sim_step
 
-        # the available_routes variable contains a dictionary of routes
-        # vehicles can traverse; to be used when routes need to be chosen
```
eugenevinitsky: where has this moved to?

flow/core/kernel/kernel.py
AboudyKreidieh: yup! AimsunScenario will be in a later PR though

flow/core/kernel/scenario/__init__.py
AboudyKreidieh: this is the __init__ file, we're just adding the stuff to be able to call them more easily later

flow/core/kernel/scenario/base.py
AboudyKreidieh: ya.. I have an idea of how to fix it, but it'll take some time. Will prioritize ASAP

flow/core/kernel/scenario/base.py
AboudyKreidieh: yup, tried to work it out for a bit but then gave up. If they overlap the environment will push "Bad start" and close anyways. But good call!

flow/core/kernel/scenario/base.py
AboudyKreidieh: creating an issue about this. This might be causing problems for some people too.

flow/envs/base_env.py
```diff
@@ -124,18 +118,31 @@ def __init__(self, env_params, sim_params, scenario):
...
+                        sim_params=sim_params)
+
+        # use the scenario class's network parameters to generate the necessary
+        # scenario components within the scenario kernel
+        self.k.scenario.generate_network(scenario)
 
         # initialize the simulation using the simulation kernel. This will use
         # the scenario kernel as an input in order to determine what network
         # needs to be simulated.
         self.traci_connection = self.k.simulation.start_simulation(
```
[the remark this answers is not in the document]
AboudyKreidieh: yup, in a later PR. Working towards it

flow/core/kernel/scenario/base.py
AboudyKreidieh: yup, need to review this as well. Will create a new PR about this

flow/core/kernel/scenario/base.py
AboudyKreidieh: We need to document all these things (and maybe delete bunching tbh). There is an issue about this

flow/core/kernel/scenario/traci.py
AboudyKreidieh: Nope completely copied! Don't worry about it

flow/core/kernel/scenario/traci.py
AboudyKreidieh: Maybe, but this will be more clear when we get there

flow/envs/base_env.py
AboudyKreidieh: after the kernel instantiation (a few lines down)

flow/core/kernel/scenario/traci.py
AboudyKreidieh: ya, I don't know anything about these.. there's an issue about writing tests for these. I think if we do that we will begin to result this, something I'm currently talking to Yashar about

PR #367, opened 2019-01-04
Remove abs position

AboudyKreidieh:
changes:
- removed `get_absolute_position` from the vehicles class and moved it to the environments that actually use it. This variable is essentially broken in environments that don't have a single path, so better not to treat it as a variable for every environment
- removed sorting, and left it in the environments that originally used it (i.e. the CoRL 2017 environments)


eugenevinitsky:
Question: I don't see anywhere where "sort_vehicles": True is defined and then actually used in the environment?

AboudyKreidieh:
I transformed `sorted_ids` into a property (similar to action_space and obs_space). It's called within the property now instead of in `additional_command` or `reset` for instance

eugenevinitsky:
Right, sorry, what I mean is that you'd expect it to be used in the relevant env run scripts, i.e. figure_eight, loop, multi_loop. I'm not seeing it called there?

AboudyKreidieh:
Basically I'm keeping it for the purposes of being able to reproduce old work, but we don't generally use then, hence me setting them to False everywhere

eugenevinitsky:
Oh sorry, now I understand. They're intentionally currently not active, we can just turn them on if we feel the need to reproduce old work. Okay.

PR #368, opened 2019-01-04
Example docs

### LINKED ISSUE #279: Update documentations with examples
If you look at: https://ray.readthedocs.io/en/latest/example-lbfgs.html
they have a whole list of examples illustrating what you can do with RLlib. We should provide one of these on our documentation page as well. Obviously we have tutorials but it's more compelling if you can see a couple exciting use cases and explanations of how they work.

eugenevinitsky:
- added documentation and pictures to the readthedocs describing the available environments. 
- closes #279


docs/source/examples.rst
```diff
@@ -0,0 +1,94 @@
...
+
+We provide a quick introduction to the traffic examples and tools that Flow comes pre-built with.
+For further details, please examine the documentation provided in the examples folder.
+A few points of clarification:
+
+* We discuss the "goal" of the example. By this we mean the standard reward that is built into the environment. It is simple to change the reward function if other goals are desired.
+
+* Occasionally we describe environments as *fully observed*. By this we mean that all relevant pieces of information (speed, relative positions, traffic light states) of the system are available to the controller. *Partially observed* refers to only a subset being available. This is the default configuration of the environment and the set of observations can be customized.
+
+* Each environment has customizable environment/networks parameters that can be used to configure it beyond what is described here. Particularly pertinent parameters are described. Additional parameters can be found by examining the documentation in the relevant scenario and environment files.
```
AboudyKreidieh:
```suggestion
* Each environment has customizable environment/network parameters that can be used to configure it beyond what is described here. Particularly pertinent parameters are described. Additional parameters can be found by examining the documentation in the relevant scenario and environment files.
```

docs/source/examples.rst
```diff
@@ -0,0 +1,94 @@
...
+`examples/rllab/stabilizing_the_ring.py`. These control environments are partially observed:
+the autonomous vehicle only observes its own distance to the leading vehicle, its speed,
+and the speed of the leading vehicle.
+
+To make this task more difficult, the environment has a configurable parameter, `ring_length`, which
+can be set to a list containing the minimum and maximum ring-size. The autonomous vehicle must
+learn to distinguish these scenarios from each other and pick the appropriate driving behavior.
+
+.. image:: ../img/stabilizing_the_ring.png
+   :width: 400
```
AboudyKreidieh: Looking at the file the pictures look huge! Part of it might be that they don't all share the same dimensions. Maybe try using `:height:` instead (if that is a thing?). Also unclear if they will look the same on readthedocs, so unsure

examples/rllib/green_wave.py
```diff
@@ -73,22 +73,22 @@ def get_non_flow_params(enter_speed, additional_net_params):
 inner_length = 800
 long_length = 100
 short_length = 800
-n = 1
-m = 5
+N_ROWS = 1
+N_COLUMNS = 5
```
AboudyKreidieh: nit: if we're changing these, should we capitalize all variables as well?

docs/source/examples.rst
eugenevinitsky: ah yeah, I actually looked at it in readthedocs, it's not too big.

PR #371, opened 2019-01-04
Upgrade autoscaler script to simplify change process

eugenevinitsky:
Previous AMI came with ray pre-built on the machine, but this would cause errors whenever Ray had a merge conflict. Now the autoscaler pip installs ray, clones our version of Ray onto the machine, and then replaces rllib and tune

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
         - cached_property
         - joblib==0.10.3
         - dill
-        - boto3=1.4.8
+        - boto3==1.4.8
         - redis~=2.10.6
         - git+https://github.com/Theano/Theano.git@adfe319ce6b781083d8dc3200fb4481b00853791#egg=Theano
         - git+https://github.com/neocxi/Lasagne.git@484866cf8b38d878e92d521be445968531646bb8#egg=Lasagne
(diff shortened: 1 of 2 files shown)


scripts/ray_autoscale.yaml
```diff
@@ -66,15 +66,20 @@ worker_nodes:
...
     # checkout your desired branch on all worker nodes
-    - cd flow && git fetch && git checkout `cat /tmp/path`
-    - cd ray && git fetch eugene_upstream && git checkout eugene_upstream/master && git pull eugene_upstream master
+    - cd flow && git fetch && git checkout master && git pull origin master
+    # set up ray by pip installing and copying over rllib and tune folder
+    - pip install --upgrade pyzmq notebook
+    - pip install ray==0.6.1
+    - pip install lz4
+    - git clone https://github.com/flow-project/ray.git
+    - cd ray && pwd && ls && git fetch && git checkout centralized_vf_pr
```
eugenevinitsky: cd ray && git fetch && git checkout master

environment.yml
```diff
@@ -19,6 +19,7 @@ dependencies:
         - cached_property
         - joblib==0.10.3
         - dill
+        - boto3=1.4.8
```
AboudyKreidieh:
```suggestion
        - boto3==1.4.8
```

AboudyKreidieh: I think, right?

PR #376, opened 2019-01-09
Coverage cleanup

AboudyKreidieh:
changes:
- removed features that are unused and conflict in some way with Aimsun
- some cleanup to the Loop merge environment
- created a `FatalFlowError` exception that mimics sumo's `FatalTraCIError` 
- added test to ensure that when not enough vehicles emerge in the network an error is raised


fywu85:
LGTM!

PR #396, opened 2019-02-01
rename "no_collide" -> "obey_safe_speed" and then move "no_collide" = 7

### LINKED ISSUE #318: rename "no_collide" -> "obey_safe_speed" and then move "no_collide" = 7

AboudyKreidieh:
resolves #318

### DIFF
--- a/examples/rllab/cooperative_merge.py
+++ b/examples/rllab/cooperative_merge.py
@@ -51,7 +51,7 @@ def run_task(*_):
         car_following_params=SumoCarFollowingParams(
             minGap=0.01,
             tau=0.5,
-            speed_mode="no_collide"
+            speed_mode="obey_safe_speed"
         ),
         lane_change_params=SumoLaneChangeParams())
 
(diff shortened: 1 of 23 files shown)


fywu85:
It seems the `no_collide` has been changed from 0b00001 to 0b00111. Since this may potentially cause confusion for existing users, consider deprecating the name `no_collide` all together. Instead, perhaps consider using something like `run_red_light` for 0b00111 (as used by the [official documentation](http://sumo.dlr.de/wiki/TraCI/Change_Vehicle_State#speed_mode_.280xb3.29)). It would be even nicer if we can throw a deprecation warning when we find `speed_mode` is set to `no_collide`. Overall, nice PR!

AboudyKreidieh:
Run red light does not denote the behavior people are supposed to achieve when using that speed mode, and I think this PR is simply supposed to make no_collide function the way people expect it to. Personally I would argue to keep as is

fywu85:
No strong objection to merge as it is!

PR #414, opened 2019-02-17
modified driving behavior by our controllers

AboudyKreidieh:
- at junctions, the car following behavior is dictated by the simulator instead of us
- if a car just enters, in new sumo it may not be subscribed. Added a fix to support new sumo in the near future

Note: these features are especially helpful when using the minicity scenario


fywu85:
Cares need to be taken to override this PR if one tries to write a dedicated controller for intersection. But this looks good for now.

PR #427, opened 2019-02-27
solve the issue with minicity router 9 ( deleted it), Added new nodes…

lucfisc:
solve the issue with minicity router 9 ( deleted it), Added new node and edges in minicity scenario


crystalzyan:
@lucfisc Lucas, The files you modified are the old examples/sumo/minicity.py and the Minicity scenario file. We actually weren't using Minicity Router 9 anymore anyways, we're using examples/sumo/minicity_subroutes2.py  instead with our own custom router included in this subroutes2 file.

crystalzyan:
I'm merging this PR since you say Xiao @yunerzxy is no longer working on Minicity, but as she was the one who created Minicity she's the best authority as to whether the new nodes and edges might break anything or not.

PR #439, opened 2019-03-03
bug fix to number of vehicles

### LINKED ISSUE #387: self.k.vehicle.num_vehicles can go negative
If you run velocity_bottleneck.py with reset_inflows turned on, after one rollout self.k.vehicle.num_vehicles is negative.

AboudyKreidieh:
It seems that vehicles are sometimes removed twice during reset. The number of vehicles in the list is not affected though, so calculating the length now instead of decrementing the values.

resolves #387


fywu85:
LGTM. Nice PR

PR #457, opened 2019-03-12
added horizon to termination in Env

### LINKED ISSUE #442: Env cannot signal the end of a run.
**Choose an option**
This is a: 
- [ ] Question
- [x] Feature request
- [ ] Bug report

**Describe the issue**
Environments that inherit from Env, should be able to indicate that they have reached the end of their simulation time line. 
For instance, consider a situation in which I am simulating the peak hour. After that period is over, I do not want to continue simulation, and intend to raise "done" in the return parameters of gym environment's step.
As it stands, Env has decided not to let me do it, and return done as true only in the case of a crash.

AboudyKreidieh:
also updated the default horizon in `EnvParams` to be infinity so that it by default does not affect anything

resolves #442

### DIFF
--- a/flow/core/params.py
+++ b/flow/core/params.py
@@ -587,7 +587,7 @@ class EnvParams:
 
     def __init__(self,
                  additional_params=None,
-                 horizon=500,
+                 horizon=float('inf'),
                  warmup_steps=0,
                  sims_per_step=1,
                  evaluate=False):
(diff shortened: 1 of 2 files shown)


fywu85:
What is the rationale behind defaulting horizon to infinity?

AboudyKreidieh:
this is so that it can be backwards compatible, in case people still want to specify the horizon from `num_steps` in the `Experiment.run(...)`

PR #486, opened 2019-03-28
tutorial for osm

AboudyKreidieh:
- added a tutorial on importing networks from openstreetmap
- defaulted routes to consist of the current edge only (to make running osm files easier)
- updated the tutorials README
- added test for default scenario routes

PR #504, opened 2019-04-17
Aimsun-Flow interface

nathanlct:
do not merge; cleaning ongoing


flow/scenarios/base_scenario.py
```diff
@@ -109,7 +109,11 @@ def __init__(self,
             self.connections = None
 
         # specify routes vehicles can take
-        self.routes = self.specify_routes(net_params)
```
AboudyKreidieh: i think we will eventually want to add routes via `specify_routes`. Can we just ignore this? As of #486 this should be producing default values, and should not break anything.

flow/core/kernel/scenario/aimsun.py
```diff
@@ -50,7 +50,12 @@ def generate_network(self, scenario):
             "osm_path": scenario.net_params.osm_path,
             'render': self.sim_params.render,
             "sim_step": self.sim_params.sim_step,
-            "traffic_lights": None
+            "traffic_lights": None,
+            "scenario_name": self.sim_params.scenario_name,
```
AboudyKreidieh: how is this different from `scenario.name` or `scenario.orig_name`?

flow/core/kernel/scenario/aimsun.py
```diff
@@ -73,39 +78,80 @@ def generate_network(self, scenario):
...
+                self._junction_list = list(
+                    set(self._edges.keys()) - set(self._edge_list))
+
+            else:
+                # load scenario from template
+                scenario_file = "flow/core/kernel/scenario/scenario_data.json"
+                filepath = os.path.join(config.PROJECT_PATH, scenario_file)
+
+                while not os.path.exists(filepath):
+                    time.sleep(0.5)
```
AboudyKreidieh: why do we want 0.5 s here? does the file take that long to create? maybe we can try something smaller, like 0.1s

flow/core/kernel/scenario/aimsun.py
```diff
@@ -73,39 +78,80 @@ def generate_network(self, scenario):
...
+                time.sleep(3.0)
+
+                with open(filepath) as f:
+                    time.sleep(0.5)
+                    content = json.load(f)
+
+                self._edges = content['sections']
+                self._edge_list = self._edges.keys()
+                self._junction_list = content['turnings']
+                # TODO load everything that is in content into the scenario
```
AboudyKreidieh: what still needs to be done for this TODO? Can we list them (will help out for later)

flow/scenarios/base_scenario.py
nathanlct: indeed it is not necessary anymore, I removed it.

flow/core/kernel/scenario/aimsun.py
nathanlct: fixed in b524edc

flow/core/kernel/scenario/aimsun.py
nathanlct: `scenario_name` is the name of the scenario used by Aimsun, eg `Dynamic Scenario 917`. When creating a template we would set this equal to `scenario.name`, but when loading one, it can have several scenarios one of which we need to choose.

flow/core/kernel/scenario/aimsun.py
nathanlct: We have the following information we can make use of (and we could possibly get more depending on needs):

- centroids (input or output)
- connections (junctions between centroids and sections)
- nodes (ie intersection, list of turnings)
- sections (length, max speed, number of lanes)
- turnings (origin and destination sections, origin and destination lanes, length, max speed)
(+ the names and IDs of everything)

The length of sections and turnings are approximations; the only thing I found to make it more precise would be to manually compute it from the raw polygon. 

We already have the edges (sections). We would need to load the centroids and connections, that could work as inflow/outflow somehow (although there are no real inflows in Aimsun). 

Then there are the junctions, defined in the rest of the code as `self._junction_list = list(set(self._edges.keys()) - set(self._edge_list))`, so just a list of section IDs, while what we retrieve is a list of origin lane/destination lane pairs -> we probably need to adapt the code so it can handle the two types of junctions, or directly upgrade junctions from a list of IDs to a list of pair of lanes (or change the structure even deeper and have for each lane/section a list of the possible next lane/section destinations).

PR #505, opened 2019-04-19
Wave atten env

AboudyKreidieh:
Made several changes to undo some regression in the stabilizing the ring experiment.

changes:
- added a `clip_actions` attribute to `EnvParams` that specifies whether to clip actions from the policy by their range when they are inputted to the reward function. Note that the actions are still clipped before they are provided to `apply_rl_actions`. From my understanding, I believe the clipping is causing the gradients to go flat in the stabilizing the ring experiment.
- modified the close attribute in the base scenario kernel, which seems to break the stabilizing the ring environment (at least when I try to run it)
- updated one of the weights in the environment to better match my memory of old learning curves.
- added a test to the reward function in the wave attenuation environment

Note: before and after training curves to come soon


AboudyKreidieh:
wave_atten_learning_curve

Learning curve for stabilizing_the_ring with new modifications

PR #521, opened 2019-05-14
Better name for 'self.last_change' array

ashkan-software:
-Split the 2D array self.last_change[num_traffic,3], into 3 arrays, that are described by better names
-Updated the file green_wave_env with the changes
-Updated the tutorial for traffic lights with the changes

Previously, the array self.last_change contained 3 (quite distinct information) about each node: (1) which direction traffic is flowing, (2) time elapsed since the last change, and (3) whether or not the traffic light is currently displaying yellow. The name 'last_change" does not describe all this information.

### DIFF
--- a/flow/envs/green_wave_env.py
+++ b/flow/envs/green_wave_env.py
@@ -1,6 +1,5 @@
 import numpy as np
 import re
-import ipdb
 
 from gym.spaces.discrete import Discrete
 from gym.spaces.box import Box
--- a/flow/envs/green_wave_env.py
+++ b/flow/envs/green_wave_env.py
@@ -1,5 +1,6 @@
 import numpy as np
 import re
+import ipdb
 
 from gym.spaces.discrete import Discrete
 from gym.spaces.box import Box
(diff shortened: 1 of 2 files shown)


AboudyKreidieh:
@kjang96 would you be able to cover this PR? It covers code you mostly worked on.

AboudyKreidieh:
@ashkan-software can you deal with the pep8 bugs:

```
./flow/envs/green_wave_env.py:91:80: E501 line too long (196 > 79 characters)
        # keeps track of the last time the traffic lights in an intersection were allowed to change (the last time the lights were allowed to change from a red-green state to a red-yellow state.).
                                                                               ^
./flow/envs/green_wave_env.py:93:80: E501 line too long (193 > 79 characters)
        # keeps track of the direction of the intersection (the direction that is currently being allowed to flow. 0 indicates flow from top to bottom, and 1 indicates flow from left to right.)
                                                                               ^
./flow/envs/green_wave_env.py:95:80: E501 line too long (137 > 79 characters)
        # value of 0 indicates that the intersection is in a red-yellow state. 1 indicates that the intersection is in a red-green state.
```

kjang96:
Hi @ashkan-software, I'm stepping through the code with ipdb breakpoints in the _apply_rl_actions function. At some points, the values of self.last_change, self.direction, self.currently_yellow are not what I expect them to be. For example, self.currently_yellow is 0 (red-yellow state) when the renderer is in a red-green state. Can you check on this? Thanks!

ashkan-software:
Apologies for the issue. I had accidentally deleted "}," in the tutorial, and that had caused the issue. I fixed the issue now and tested and it works. I hope this works (after all this headache for you!)

PR #524, opened 2019-05-18
New router

AboudyKreidieh:
This PR introduces a new method for specifying multiple routes from a given starting edge. Routes now can be specified in one of three ways (as documented in the the scenario class):

- In this case of deterministic routes (as is the case in the ring road scenario), the routes can be specified as dictionary where the key element represents the starting edge and the element is a single list of edges the vehicle must traverse, with the first edge corresponding to the edge the vehicle begins on. Note that the edges must be connected for the route to be valid. For example (from flow/scenarios/loop.py):

```python
def specify_routes(self, net_params):
    return {
        "top": ["top", "left", "bottom", "right"],
        "left": ["left", "bottom", "right", "top"],
        "bottom": ["bottom", "right", "top", "left"],
        "right": ["right", "top", "left", "bottom"]
    }
```

- Alternatively, if the routes are meant to be stochastic, each element can consist of a list of (route, probability) tuples, where the first element in the tuple is one of the routes a vehicle can take from a specific starting edge, and the second element is the probability that vehicles will choose that route. Note that, in this case, the sum of probability values for each dictionary key must sum up to one. For example, modifying the code snippet we presented above, another valid way of representing the route in a more probabilistic setting is:

```python
def specify_routes(self, net_params):
    return {
        "top": [(["top", "left", "bottom", "right"], 1)],
        "left": [(["left", "bottom", "right", "top"], 1)],
        "bottom": [(["bottom", "right", "top", "left"], 1)],
        "right": [(["right", "top", "left", "bottom"], 1)]
    }
```

- Finally, if you would like to assign a specific starting edge and route to a vehicle with a specific ID, you can do so by adding a element into the dictionary whose key is the name of the vehicle and whose content is the list of edges the vehicle is meant to traverse as soon as it is introduced to the network. As an example, assume we have 4 vehicles named 'human_0', 'human_1', 'human_2', and 'human_3'. Then, an appropriate definition of the routes may look something like:

```python
def specify_routes(self, net_params):
    return {
        "human_0": ["top", "left", "bottom", "right"],
        "human_1": ["left", "bottom", "right", "top"],
        "human_2": ["bottom", "right", "top", "left"],
        "human_3": ["right", "top", "left", "bottom"]
    }
```


flow/scenarios/base_scenario.py
```diff
@@ -548,14 +548,80 @@ def specify_connections(self, net_params):
...
+          element in the tuple is one of the routes a vehicle can take from a
+          specific starting edge, and the second element is the probability
+          that vehicles will choose that route. Note that, in this case, the
+          sum of probability values for each dictionary key must sum up to one.
+
+          For example, modifying the code snippet we presented above, another
+          valid way of representing the route in a more probabilistic setting
+          is:
+
+          >>> def specify_routes(self, net_params):
```
fywu85: Can you give a more representative example of this case? Perhaps something like:
```
"top-random": [
    (["top", "left", "bottom", "right"], 0.9)
    (["top", "right", "bottom", "left"], 0.1)
    ],
```

flow/scenarios/base_scenario.py
```diff
@@ -548,14 +548,80 @@ def specify_connections(self, net_params):
...
+          route to a vehicle with a specific ID, you can do so by adding a
+          element into the dictionary whose key is the name of the vehicle and
+          whose content is the list of edges the vehicle is meant to traverse
+          as soon as it is introduced to the network.
+
+          As an example, assume we have 4 vehicles named 'human_0', 'human_1',
+          'human_2', and 'human_3'. Then, an appropriate definition of the
+          routes may look something like:
+
+          >>> def specify_routes(self, net_params):
```
fywu85: What is differentiating this use case with the first one? Maybe I have missed it but how is it decided which way the `specify_routes` method will be parsed?

nathanlct:
@AboudyKreidieh Is it possible to merge this? (I can help with this PR if needed) It would be super useful to have non-deterministic routes for the i-210!

flow/scenarios/base_scenario.py
AboudyKreidieh: done

flow/scenarios/base_scenario.py
AboudyKreidieh: there is a comment in the end about it

PR #531, opened 2019-06-02
Scripting API for Aimsun

nathanlct:
## Pull request information

- **Status**: Ready to merge
- **Kind of changes**: API to script with Aimsun
- **Code affected**: `utils/aimsun`

## Description

Made an API to do Aimsun scripting from Flow in an easier and more pythonic way.

I also added back the `Aimsun_Flow.ang` file used when generating templates, which apparently got deleted at some point.


.coveragerc
```diff
@@ -12,6 +12,7 @@ omit =
     *aimsun/run.py
     *aimsun/api.py
     *aimsun/generate.py
+    *aimsun/load.py
```
AboudyKreidieh: Good call on this. But we should begin thinking about how we are going to start testing aimsun features. I've been thinking about generating dummy modules to replace the aimsun module's in tests, but we may need to figure out more clever ways as well.

tests/fast_tests/test_aimsun_scripting_api.py
```diff
@@ -0,0 +1,313 @@
+import os
+import unittest
+
+from flow.utils.aimsun.scripting_api import AimsunTemplate
+
+os.environ['TEST_FLAG'] = 'True'
+
+
+class TestGUISystemBase(object):
```
AboudyKreidieh: Nice! This is what I was thinking about

flow/utils/aimsun/scripting_api.py
```diff
@@ -0,0 +1,262 @@
...
+
+        In cases (2) and (3), see 'save' to then save the template.
+
+        This class takes as parameter the two high-level objects provided
+        when interfacing with Aimsun: GKSystem and GKGUISystem. After having
+        imported this class, you should be able to create an AimsunTemplate
+        object as follows:
+
+            model = AimsunTemplate(GKSystem, GKGUISystem)
+        """
```
AboudyKreidieh: can you add the parameters to the docstring?

flow/utils/aimsun/scripting_api.py
```diff
@@ -0,0 +1,262 @@
...
+        - exp.setDataValue(model.getColumn(...), ...) becomes
+          exp.set_data_value(model.get_column(...), ...)
+        - s.setName(new_name) becomes s.name = new_name
+        etc.
+
+        This method directly modifies the object and does not return anything.
+
+        For back-compatibility, it is still possible to call the original
+        Aimsun methods.
+        """
```
AboudyKreidieh: add parameters to docstring

flow/utils/aimsun/scripting_api.py
```diff
@@ -0,0 +1,262 @@
...
+
+    ####################################################################
+    #                Methods to retrieve Aimsun objects                #
+    ####################################################################
+
+    def __getattr__(self, name):
+        """If trying to access an attribute in this AimsunTemplate object
+        fails, try to access it into the Aimsun model object
+        """
+        return getattr(self.model, name)
```
AboudyKreidieh: add parameters to docstring

.coveragerc
nathanlct: I'll look into it! By the way, why is `*aimsun/api.py` in this file if we have a `test_aimsun_api.py` file?

AboudyKreidieh: mm, ya, I think that script doesn't actually cover the full API, but uses a dummy instead, and just checks stuff like the aimsun objects we created. Something to look to in the future as well.

PR #541, opened 2019-06-07
Remove obsolete features from bottleneck environment

nathanlct:
## Pull request information

- **Status**: Ready to merge
- **Kind of changes**: Cleaning
- **Code affected**: Bottleneck env

---

## Description

Cleaned bottleneck env for Flow release 0.4.0. 

- removed the function `get_avg_bottleneck_velocity` that is not used anywhere
- removed the function `distance_to_bottleneck` that is not used anywhere

### DIFF
--- a/flow/envs/bottleneck_env.py
+++ b/flow/envs/bottleneck_env.py
@@ -111,7 +111,7 @@ class BottleneckEnv(Env):
         super().__init__(env_params, sim_params, scenario, simulator)
         env_add_params = self.env_params.additional_params
         # tells how scaled the number of lanes are
-        self.scaling = scenario.net_params.additional_params.get("scaling")
+        self.scaling = scenario.net_params.additional_params.get("scaling", 1)
         self.edge_dict = defaultdict(list)
         self.cars_waiting_for_toll = dict()
         self.cars_before_ramp = dict()
--- a/flow/envs/bottleneck_env.py
+++ b/flow/envs/bottleneck_env.py
@@ -112,7 +112,7 @@ class BottleneckEnv(Env):
         env_add_params = self.env_params.additional_params
         # tells how scaled the number of lanes are
         self.scaling = scenario.net_params.additional_params.get("scaling", 1)
-        self.edge_dict = defaultdict(list)
+        self.edge_dict = dict()
         self.cars_waiting_for_toll = dict()
         self.cars_before_ramp = dict()
         self.toll_wait_time = np.abs(
@@ -137,7 +137,6 @@ class BottleneckEnv(Env):
                                       cycle_offset * self.scaling * MAX_LANES,
                                       self.scaling * MAX_LANES)
         self.green_time = 4
-        self.red_min = 2
         self.feedback_coeff = env_add_params.get("feedback_coeff", 20)
 
         self.smoothed_num = np.zeros(10)  # averaged number of vehs in '4'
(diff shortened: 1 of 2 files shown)


flow/envs/bottleneck_env.py
```diff
@@ -484,14 +454,10 @@ def get_state(self):
...
             num_lanes = MAX_LANES * self.scaling
-            headway = np.asarray([1000
-                                  for _ in range(num_lanes)]) / headway_scale
-            tailway = np.asarray([1000
-                                  for _ in range(num_lanes)]) / headway_scale
-            vel_in_front = np.asarray([0 for _ in range(num_lanes)
-                                       ]) / self.max_speed
-            vel_behind = np.asarray([0 for _ in range(num_lanes)
-                                     ]) / self.max_speed
+            headway = np.asarray([1000] * num_lanes) / headway_scale
```
AboudyKreidieh: I always get nervous using `[...] * ...` because it creates copies of the same variable instead of generating new ones in memory. Maybe switch to:
```
headway = np.ones(num_lanes) * 1000 / headway_scale
```
same for the three variables below.

flow/envs/bottleneck_env.py
```diff
@@ -111,8 +110,8 @@ def __init__(self, env_params, sim_params, scenario, simulator='traci'):
         super().__init__(env_params, sim_params, scenario, simulator)
         env_add_params = self.env_params.additional_params
         # tells how scaled the number of lanes are
-        self.scaling = scenario.net_params.additional_params.get("scaling")
-        self.edge_dict = defaultdict(list)
+        self.scaling = scenario.net_params.additional_params.get("scaling", 1)
+        self.edge_dict = dict()
```
AboudyKreidieh: is there a reason dict's are preferable here instead of defaultdict's? Just curious, I'm open to the change

flow/envs/bottleneck_env.py
```diff
@@ -642,8 +608,7 @@ def __init__(self, env_params, sim_params, scenario, simulator='traci'):
 
         # default (edge, segment, controlled) status
         add_env_params = self.env_params.additional_params
-        default = [("1", 1, True), ("2", 1, True), ("3", 1, True),
-                   ("4", 1, True), ("5", 1, True)]
+        default = [(str(i), 1, True) for i in range(1, 6)]
```
AboudyKreidieh: Nice call!

flow/envs/bottleneck_env.py
nathanlct: There's no problem when you use `[...] * ...` with immutable types like `int` since you can never modify their value so it doesn't matter that they're stored at the same location ([link](https://stackoverflow.com/a/24557558/11584932), [link](https://stackoverflow.com/a/983727/11584932))

flow/envs/bottleneck_env.py
nathanlct: Because it wasn't using the 'default' feature of the type anywhere (plus it was the only `defaultdict` in the file so I just removed the `defaultdict` import)

AboudyKreidieh: sounds good, thanks for catching that

flow/envs/bottleneck_env.py
AboudyKreidieh: awesome, I'm convinced

PR #557, opened 2019-06-21
Modified the general structure around the "emission to csv function"

alt113:
## Pull request information

- **Status**: Ready to merge
- **Kind of changes**: This PR introduces a general fix around the use of the 'emission to csv' function and adds some flexibility to it in the sense that you can now specify which characteristic emissions to monitor instead of having all of them which creates a more stable csv file especially when running a large timestep monitoring experiment over many vehicles.
- **Code affected**: The run function of the Environment class now has an optional 'args' function that takes in the variable number of characteristic emissions to monitor in the new emission to csv function. The utils file is where the 'generic_observations' and 'new_emission_to_csv' functions reside. The old 'emission_to_csv' function is still available in the utils.py file.

---

## Description

This PR introduces a general fix around the use of the 'emission to csv' function and adds some flexibility to it in the sense that you can now specify which characteristic emissions to monitor instead of having all of them which creates a more stable csv file especially when running a large timestep monitoring experiment over many vehicles.

Features:
- more flexibility around emission monitoring 
- more stable output files (in terms of tractability in size)


AboudyKreidieh:
this is stale. we can come back around if we need it later

Issue #551, opened 2019-06-18
Incompatibilities with Aimsun in new Flow release


nathanlct:
**Choose an option**
This is a: 
- [ ] Question
- [ ] Feature request
- [x] Bug report

**Describe the issue**
Aimsun examples are not running in Flow release 0.4.0

Two bugs I had but there might be more, I didn't dwelve into it much:

- There is a new method `get_initial_speed` in `vehicle/traci.py` that is not in `vehicle/base.py` nor in `vehicle/aimsun.py`, but at some point it gets called on the Aimsun vehicle kernel, thus resulting in an error.

- There is a new section in `aimsun/vehicle.py` to add vehicles (on top of the `add` and `add_vehicle` methods) that doesn't define everything the vehicle needs (in particular `'type_name'`) 

```python
        self.__vehicles.clear()
        for typ in vehicles.initial:
            for i in range(typ['num_vehicles']):
                veh_id = '{}_{}'.format(typ['veh_id'], i)
                self.__vehicles[veh_id] = dict()
                self.__vehicles[veh_id]['type'] = typ['veh_id']
                self.__vehicles[veh_id]['initial_speed'] = typ['initial_speed']
                self.num_vehicles += 1
                if typ['acceleration_controller'][0] == RLController:
                    self.num_rl_vehicles += 1
```

I'll look into it more later, especially I'll try to find a way to write tests for Aimsun so that this doesn't happen in the future. 

**To Reproduce**
Run an Aimsun example (sugiyama, figure eight...) with the latest Flow release.

nathanlct:
Fixed in #565

PR #565, opened 2019-06-25
Fixed Aimsun bugs

Yasharzf:
## Pull request information

- **Status**: Ready to merge / In development
- **Kind of changes**: ?
- **Code affected**: ?
- **Related PR or issue**: (#551)

---

## Description

Fixed Aimsun bugs
- changed sim step from .1 to .5 since Aimsun can not handle sim_step = 0.1
- fixed bugs in Aimsun vehicle file.  added get_initial_speed function
- fixed a bug in the generator file. Append path should be before the imports

Features:
- 
- 
-


examples/aimsun/stabilizing_the_ring.py
```diff
@@ -17,7 +17,7 @@
 
 def run_task(*_):
     """Implement the run_task method needed to run experiments with rllab."""
-    sim_params = SumoParams(sim_step=0.1, render=False, seed=0)
+    sim_params = SumoParams(sim_step=0.5, render=False, seed=0)
```
AboudyKreidieh: replace with `AimsunParams`

flow/core/kernel/vehicle/aimsun.py
```diff
@@ -119,15 +119,16 @@ def initialize(self, vehicles):
...
                     self.num_rl_vehicles += 1
 
-        for tracked_type in self.tracked_vehicle_types:
-            self.num_type[tracked_type] = 0
-            self.total_num_type[tracked_type] = 0
-            self.type_parameters[tracked_type] = {}
+        # for tracked_type in self.tracked_vehicle_types:
+        #     self.num_type[tracked_type] = 0
+        #     self.total_num_type[tracked_type] = 0
+        #     self.type_parameters[tracked_type] = {}
```
AboudyKreidieh: why do we need to comment these lines? not disapproving, just curious

examples/aimsun/stabilizing_the_ring.py
Yasharzf: Good catch!

flow/core/kernel/vehicle/aimsun.py
Yasharzf: The bug here is that it assigns an empty dict to type_parameters. Since I commented these lines, now the load.py doesn't work. Nathan is working on it.

Issue #543, opened 2019-06-10
Visualization - Plotting Reward


KidTrunks:
![Internal Server error](https://user-images.githubusercontent.com/45654203/59170054-29827100-8b80-11e9-96e6-7654fd7cca42.png)
**This is a:**
Question

**Describe the Issue:**
Hi I am trying to visualize Automated vehicle learning by using the plotting reward function. The simulation has run to it's completion and I made sure the progress.csv and .json files are in the folder. The .json file was variant.json instead of params.json. 

However after clicking _View http://localhost:5000 in your browser,_ 
I encountered 500 internal server error and the following error code:

(flow) python@python-VirtualBox:~/rllab-multiagent/rllab/viskit$ python frontend.py /home/python/rllab-multiagent/data/local/stabilizing-the-ring/stabilizing-the-ring_2019_05_19/
Importing data from ['/home/python/rllab-multiagent/data/local/stabilizing-the-ring/stabilizing-the-ring_2019_05_19/']...
Reading /home/python/rllab-multiagent/data/local/stabilizing-the-ring/stabilizing-the-ring_2019_05_19/progress.csv
Reading /home/python/rllab-multiagent/data/local/stabilizing-the-ring/stabilizing-the-ring_2019_05_19/gym_log/progress.csv
Done! View http://localhost:5000 in your browser
 * Serving Flask app "frontend" (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
 * Running on http://0.0.0.0:5000/ (Press CTRL+C to quit)
AverageReturn None None None
[]
[]
[]
[2019-06-10 12:42:48,202] ERROR in app: Exception on / [GET]
Traceback (most recent call last):
  File "/home/python/anaconda3/envs/flow/lib/python3.5/site-packages/flask/app.py", line 2311, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/python/anaconda3/envs/flow/lib/python3.5/site-packages/flask/app.py", line 1834, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/python/anaconda3/envs/flow/lib/python3.5/site-packages/flask/app.py", line 1737, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/python/anaconda3/envs/flow/lib/python3.5/site-packages/flask/_compat.py", line 36, in reraise
    raise value
[...]

PR #574, opened 2019-06-28
Updated environment.yml

kjang96:
## Pull request information

- **Status**: Ready to merge
- **Kind of changes**: Updated environment.yml to modify the pre-existing version of pyplot; also adds matplotlib to dependencies. 
- **Code affected**: only environment.yml
- **Related PR or issue**: see https://github.com/flow-project/flow/issues/543 for why we need a pyplot update

---

## Description

 Updated environment.yml to modify the pre-existing version of pyplot; also adds matplotlib to dependencies. Is there a reason why requirements.txt still exists? We should just bring all the requirements.txt packages into environment.yml

Features:
- 
- 
-


cathywu:
I still find the requirements.txt useful for creating our AMI instances and
patching partial builds. I believe that the yaml approach has the issue of
failing silently / prematurely, which leads to a partial install. In that
case I often use the requirements file to complete the install.
-- 
Best wishes,
Cathy

AboudyKreidieh:
this PR is redundant with #758 and the tests here are failing, so closing

PR #575, opened 2019-06-28
Simplify grid scenario

nathanlct:
## Pull request information

- **Status**: Ready to merge
- **Kind of changes**: Improved code & made it more readable
- **Code affected**: Grid scenario (& a few places that use it)

---

## Description

Went through all the grid scenario code and remade it in a more concise and understandable way (especially the inner/outer nodes/edges building). 

**Todo:**
**- is the method `specify_edge_starts` necessary? no one seems to know how it works and I don't see it being used anywhere (which is true for a lot of scenarios actually).**
**- also in all scenarios, the `gen_custom_start_pos` method is marked `@staticmethod`: is this necessary? I don't see that feature being used either.**


AboudyKreidieh:
So `specify_edge_starts` is not necessary, but it's used by two old experiments. I'm phasing it out. A PR is coming up to remove `specify_internal_edgestarts`, and it will be followed by one to remove `specify_edgestarts`

PR #578, opened 2019-07-01
Remove no_internal_links

### LINKED ISSUE #489: remove no_internal_links
This feature is sumo-specific and only needed by the loop scenario. Would be better to remove it.

AboudyKreidieh:
## Pull request information

- **Status**: Ready to merge
- **Kind of changes**: Cleanup
- **Code affected**: scenario and net_params
- **Related PR or issue**: resolves #489 

---

## Description

The `no_internal_link` attribute in `NetParams` exists only for the ring road (all other networks have junctions) and is only valid in sumo. Accordingly, this PR serves to remove it. The changes are as follows:

-  removed mentions of `no_internal_links` 
- added internal edge starts to loops
- modified tests to match changes


AboudyKreidieh:
@ashkan-software can you review this? I need to merge this in to change some of the examples stuff without encountering merge conflicts later?

ashkan-software:
I can. I am just moving out of my apartment and can get back to this on Monday. Is that too late?

flow/benchmarks/figureeight2.py
```diff
@@ -63,7 +63,6 @@
     # network-related parameters (see flow.core.params.NetParams and the
     # scenario's documentation or ADDITIONAL_NET_PARAMS component)
     net=NetParams(
-        no_internal_links=False,
```
eugenevinitsky: Okay, my one comment is, if something is the default for every single file, it should actually just be set as a default in the Class definition. This would have made it a lot easier to remove all of these.

flow/scenarios/loop.py
```diff
@@ -196,7 +195,20 @@ def specify_edge_starts(self):
         """See parent class."""
         edgelen = self.net_params.additional_params["length"] / 4
 
-        edgestarts = [("bottom", 0), ("right", edgelen), ("top", 2 * edgelen),
-                      ("left", 3 * edgelen)]
+        edgestarts = [("bottom", 0),
```
eugenevinitsky: Magic numbers! No magic numbers without at least a comment explaining why they're there (or alternately, a name that explains why they're there)

flow/scenarios/loop.py
```diff
@@ -196,7 +195,20 @@ def specify_edge_starts(self):
...
+                      ("right", edgelen + 0.1),
+                      ("top", 2 * edgelen + 0.2),
+                      ("left", 3 * edgelen + 0.3)]
+
+        return edgestarts
+
+    def specify_internal_edge_starts(self):
+        """See parent class."""
+        edgelen = self.net_params.additional_params["length"] / 4
```
eugenevinitsky: Same as above. Not really fair to through the division by 4 here and above on you, but since you're making the changes anyways, it'd be good to deal with that magic number as well.

flow/visualize/time_space_diagram.py
```diff
@@ -247,14 +247,17 @@ def _ring_road(data, params, all_time):
...
+    edgelen = params['net'].additional_params["length"] / 4
 
-    # generate edge starts
     edgestarts = {
-        'bottom': 0,
-        'right': total_len / 4,
-        'top': total_len / 2,
-        'left': 3 * total_len / 4
+        "bottom": 0,
+        ":right_0": edgelen,
```
eugenevinitsky: Maaaaagic numbers. Same as above.

tests/fast_tests/test_controllers.py
```diff
@@ -521,7 +521,7 @@ def test_find_intersection_dist(self):
             for veh_id in ids
         ]
 
-        expected = [-10, 0, 23., 34.5, 46.]
+        expected = [-10, 0, 23.1, 34.7, 46.3]
```
eugenevinitsky: In this case, the magic numbers are okay.

tests/fast_tests/test_scenario_base_class.py
```diff
@@ -823,44 +820,32 @@ class TestNextPrevEdge(unittest.TestCase):
...
-        Tests the prev_edge() method in the presence of internal links.
-        """
-        env, scenario = figure_eight_exp_setup()
         prev_edge = env.k.scenario.prev_edge("bottom", 0)
         expected_prev_edge = [(':bottom_0', 0)]
-
         self.assertCountEqual(prev_edge, expected_prev_edge)
 
-    def test_next_edge_no_internal_links(self):
+    def test_next_prev_edge_ring_road(self):
```
eugenevinitsky: I think this could be named slightly better as it is it reads like it tests the "next previous edge" method. Maybe "test_edge_finding"?

flow/benchmarks/figureeight2.py
AboudyKreidieh: agreed

PR #623, opened 2019-07-12
Multi-agent traffic light grid

cathywu:
## Pull request information

- **Status**: ready for review
- **Kind of changes**: New multi-agent environment.
- **Related PR or issue**: N/A

## Description

- Introduces multi-agent version of `PO_TrafficLightsGridEnv`.
- Includes some updated, revised, and cleaned up documentation relating to multi-agent Flow.
- Includes parameterized run script, which supports different grid sizes, local vs cluster runs, different vehicle inflow rates.

PR #629, opened 2019-07-12
Fix vehicle accounting (for multi-agent settings)

cathywu:
## Pull request information

- **Status**: ready to merge
- **Kind of changes**: bug fix, for visualization of multi-agent experiments
- **Related PR or issue**: relies on PR #623

## Description

In multi-agent settings, the same vehicles can be added multiple times by the Flow kernel due to being observed by multiple agents and perhaps other reasons as well. These vehicles were not being properly removed, so SUMO and Flow were crashing when trying to access the vehicles (for visualization purposes) that no longer existed in the system. This PR properly ensures that vehicles are added to the appropriate lists just once.

This PR includes some minor refactoring of related try-except statements. In general, we should refrain from using those too much, since it may hide important bugs and code failures.


eugenevinitsky:
I'll look at this post the multiagent traffic light PR since this is a superset of that.

cathywu:
That's perfect, thanks!

flow/core/kernel/vehicle/traci.py
```diff
@@ -244,7 +244,8 @@ def _add_departed(self, veh_id, veh_type):
         if veh_type not in self.type_parameters:
             raise KeyError("Entering vehicle is not a valid type.")
 
-        self.__ids.append(veh_id)
+        if veh_id not in self.__ids:
```
eugenevinitsky: Good catch.

PR #632, opened 2019-07-13
Compute and display inflows and throughput efficiency

cathywu:
## Pull request information

- **Status**: ready to merge

## Description

Introduces an efficiency metric (between 0 and 1) which can be used to assess the optimality of our experiments. Efficiency of 1 means that the outflow (throughput) is as high as the inflow. We can't do much better than that. This is displayed when with the rllib visualizer.

### DIFF
--- a/flow/visualize/visualizer_rllib.py
+++ b/flow/visualize/visualizer_rllib.py
@@ -196,6 +196,7 @@ def visualizer_rllib(args):
 
     # Simulate and collect metrics
     final_outflows = []
+    final_inflows = []
     mean_speed = []
     std_speed = []
     for i in range(args.num_rollouts):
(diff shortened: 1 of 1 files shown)


flow/visualize/visualizer_rllib.py
```diff
@@ -239,6 +240,13 @@ def visualizer_rllib(args):
             rets.append(ret)
         outflow = vehicles.get_outflow_rate(500)
```
eugenevinitsky: Can we add a warning here if the total horizon is less than this? If it is, these numbers might be a good deal skewed.

cathywu:
So, I considered this. For an inflow rate of less than 500, what's returns is the full horizon, so there actually shouldn't be anything unexpected. Am I missing something?

What I mean is that say you have a list A = [1,2,3]. A[:-500] yields [1,2,3].

cathywu:
Merging for now so that PR #629 can be clean and ready, but we can keep discussing.

PR #648, opened 2019-07-17
Add padding and docstring to `green_wave_env.k_closest_to_intersection`

nathanlct:
## Pull request information

- **Status**: Read to merge
- **Kind of changes**: New feature & docstring
- **Related PR or issue**: -

## Description

Function `green_wave_env.k_closest_to_intersection`:
- completed docstring and added usage examples
- added an optional `padding` argument to not lose information when calling the function on several edges

### DIFF
--- a/flow/envs/green_wave_env.py
+++ b/flow/envs/green_wave_env.py
@@ -433,40 +433,40 @@ class TrafficLightGridEnv(Env):
                 pos="0",
                 speed="max")
 
-    def get_closest_to_intersection(self, edges, k, padding=False):
+    def get_closest_to_intersection(self, edges, num_closest, padding=False):
         """Return the IDs of the vehicles that are closest to an intersection.
 
-        For each edge in edges, return the ids (veh_id) of the k vehicles
-        in edge that are closest to an intersection (the intersection they
-        are heading towards).
+        For each edge in edges, return the IDs (veh_id) of the num_closest
+        vehicles in edge that are closest to an intersection (the intersection
+        they are heading towards).
 
         This function performs no check on whether or not edges are going
         towards an intersection or not, it just gets the vehicles that are
         closest to the end of their edges.
 
-        If there are less than k vehicles on an edge, the function performs
-        padding by adding empty strings "" instead of vehicle ids if the
-        padding parameter is set to True
+        If there are less than num_closest vehicles on an edge, the function
+        performs padding by adding empty strings "" instead of vehicle ids if
+        the padding parameter is set to True.
 
         Parameters
         ----------
         edges : str | str list
-            id of an edge or list of edge ids
-        k : int (> 0)
-            number of vehicles to consider on each edge
+            ID of an edge or list of edge IDs.
+        num_closest : int (> 0)
+            Number of vehicles to consider on each edge.
         padding : bool (default False)
-            if there are less than k vehicles on an edge, perform padding by
-            adding empty strings "" instead of vehicle ids if the padding
-            parameter is set to True (note: leaving padding to False while
-            passing a list of several edges as parameter can lead to
+            If there are less than num_closest vehicles on an edge, perform
+            padding by adding empty strings "" instead of vehicle ids if the
+            padding parameter is set to True (note: leaving padding to False
+            while passing a list of several edges as parameter can lead to
             information loss since you will not know which edge, if any,
-            contains less than k vehicles)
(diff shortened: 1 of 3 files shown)


flow/envs/green_wave_env.py
```diff
@@ -433,18 +433,79 @@ def _reroute_if_final_edge(self, veh_id):
...
+
+        Usage
+        -----
+        For example, consider the following network, composed of 4 edges
+        whose ids are "edge0", "edge1", "edge2" and "edge3", the numbers
+        being vehicles all headed towards intersection x. The id of the vehicle
+        with number n is "veh{n}" (edge "veh0", "veh1"...).
+
+                            edge1
+                            |   |
```
ashkan-software: This is aweosme!! Nice job Nathan!

nathanlct: Thanks Ashkan you're awesome too!!

flow/envs/green_wave_env.py
```diff
@@ -453,17 +514,19 @@ def k_closest_to_intersection(self, edges, k):
...
         if isinstance(edges, list):
             ids = [self.k_closest_to_intersection(edge, k) for edge in edges]
-            # flatten the list before returning it
+            # flatten the list and return it
             return [veh_id for sublist in ids for veh_id in sublist]
 
         # get the ids of all the vehicles on the edge 'edges' ordered by
-        # increasing distance to intersection
+        # increasing distance to end of edge (intersection)
         veh_ids_ordered = sorted(
```
cathywu: Since we sort the vehicles, it looks like the edge information is lost (which edge the vehicle was on). This appears to be true for the original method as well. While we're at it, should we also return the edge that each vehicle is associated with? And/or we can make it option to sort the vehicles. And/or we can sort the vehicles within each edge, but keep the edges list ordered as it was passed in. This last one, I feel, is the most desirable, at least for the multi-agent settings I'm thinking about.

cathywu: Actually, according to your docstring, the last and most desirable option is indeed what you intended. But the last line in this function `veh_ids_ordered[:k] + pad_lst` seems to add all the padding at the end and not per edge. Am I missing something?

flow/envs/green_wave_env.py
```diff
@@ -433,18 +433,79 @@ def _reroute_if_final_edge(self, veh_id):
...
+
+        If there are less than k vehicles on an edge, the function performs
+        padding by adding empty strings "" instead of vehicle ids if the
+        padding parameter is set to True
+
+        Parameters
+        ----------
+        edges : str | str list
+            id of an edge or list of edge ids
+        k : int (> 0)
```
cathywu: Consider renaming k --> num_neighbors or num_nearest for a more descriptive variable name.

flow/envs/green_wave_env.py
cathywu: Oh. Understood. Recursive call. Awesome.

flow/envs/green_wave_env.py
```diff
@@ -433,18 +433,79 @@ def _reroute_if_final_edge(self, veh_id):
...
+
+        >>> k_closest_to_intersection("edge0", 8)
+        ["veh6", "veh5", "veh4", "veh3", "veh2", "veh1"]
+
+        >>> k_closest_to_intersection("edge0", 8, padding=True)
+        ["veh6", "veh5", "veh4", "veh3", "veh2", "veh1", "", ""]
+
+        >>> k_closest_to_intersection(["edge0", "edge1", "edge2", "edge3"], 3,
+                                      padding=True)
+        ["veh6", "veh5", "veh4", "veh8", "veh7", "", "", "", "", "veh9",
```
cathywu: Very nice and descriptive docstring!!

flow/envs/green_wave_env.py
nathanlct: Adding padding solves the information loss, if the parameter is set to True. It might have been more convenient to return a list of list, but then it wouldn't have been backward-compatible and it might be simpler to deal with a flattened list, so I left it like this.

flow/envs/green_wave_env.py
nathanlct: Done!

flow/envs/green_wave_env.py
nathanlct: Thanks!

PR #654, opened 2019-07-19
Exception handling

cathywu:
## Pull request information

- **Status**: ready to merge
- **Kind of changes**: more descriptive error messages

## Description

I'm adding more descriptive error messages in `TraCIVehicle`, `Env`, and `MultiEnv` so that we have fewer silent failures. In general, we should limit our use of try-except.

A few locations are commented out (up for discussion), because those are somehow intentional try-except messages, and currently would generate a lot of error messages if we printed the traceback.

### DIFF
--- a/flow/core/kernel/vehicle/traci.py
+++ b/flow/core/kernel/vehicle/traci.py
@@ -215,7 +215,7 @@ class TraCIVehicle(KernelVehicle):
                 self.__vehicles[veh_id]["leader"] = headway[0]
                 try:
                     self.__vehicles[headway[0]]["follower"] = veh_id
-                except KeyError as e:
+                except KeyError:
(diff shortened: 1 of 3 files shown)


flow/core/kernel/vehicle/traci.py
```diff
@@ -214,8 +215,8 @@ def update(self, reset):
                 self.__vehicles[veh_id]["leader"] = headway[0]
                 try:
                     self.__vehicles[headway[0]]["follower"] = veh_id
-                except KeyError:
-                    pass
+                except KeyError as e:
```
AboudyKreidieh:
```suggestion
                except KeyError:
```

cathywu:
Thanks for the fix @AboudyKreidieh !

PR #664, opened 2019-07-23
[Do not merge until discussion] Remove 99.9% of rllab

eugenevinitsky:
With the exception of one file whose purpose I'm confused about (examples/aimsun/stabilizing_the_ring.py) RLlab has been removed.


AboudyKreidieh:
We don't need the aimsun file. And I'm definitely pro this is we want to do it. We can make it part of the 0.4.1 release

eugenevinitsky:
I'd like to check before we do this if it's possible to just convert our current rllab support into support for Garage. I don't think it's a super popular library but since we've built the support out anyways...

cathywu:
+1, perhaps we can sample a few folks who are using Garage and get a sense
of what their primary use cases are (vs baselines, RLlib, etc). then we can
see if the folks using Garage may benefit from Flow and vice versa.

AboudyKreidieh:
Personally am pro replacing rllab with stable-baselines more so than any other RL library. I will resolve merge conflicts in this PR and replace the remainign 0.1%, and if we are happy with this then let's feel free to merge

AboudyKreidieh:
If no one has any objections I can approve.

eugenevinitsky:
Thanks! I'll merge this simultaneous with the new RL library, so we always have an easy to use one + RLlib

eugenevinitsky:
@AboudyKreidieh, is it good to you now?

PR #666, opened 2019-07-27
Inflows (more features + new tuto)

nathanlct:
## Pull request information

- **Status**: Ready to merge
- **Kind of changes**: New inflow features, new inflows tutorial
- **Related PR or issue**: -

## Description

Link to the tuto notebook: https://github.com/flow-project/flow/blob/better-inflows/tutorials/tutorial12_inflows.ipynb

- added more parameters to `InFlows.add`, and completed the docstring (before there was just a link to sumo's wiki) -- modified the way inflows are created for sumo accordingly -- added some checks to catch easy mistakes with inflow params
- rewrote the inflows tutorial to take account of this changes and make it more understandable, and added a example that makes use of almost every relevant features of inflows
- also changed `num_vehicles`'s default value from 1 to 0 when adding vehicle types in `VehicleParams`, to make it more explicit that we can have inflows without needed to add initials vehicles to the network (btw when there's no vehicles, numpy raises a mean of empty slice warning, but it doesn't say where it comes from -- maybe create an issue for this?)


flow/core/params.py
```diff
@@ -384,6 +384,8 @@ class SimParams(object):
         specifies whether to render the radius of RL observation
     pxpm : int, optional
         specifies rendering resolution (pixel / meter)
+    color_vehicles : bool, optional
```
AboudyKreidieh: really like this feature. it'll definitely be helpful in other cases too

Issue #669, opened 2019-07-29
Mac install shell script has mismatched quotes


eugenevinitsky:
Line 4 of setup_sumo_osx.sh has mismatched quotes so that it can't actually run if you don't have brew.

eugenevinitsky:
This was resolved in #670

PR #670, opened 2019-07-29
Fix command execution in mac osx

eugenevinitsky:
Fix for #669. The OSX shell script would break if you didn't have bash rather than printing and exiting gracefully.


AboudyKreidieh:
merge with master and the build should stop erroring

PR #671, opened 2019-07-30
Multi-agent training on highway with ramps

nathanlct:
## Pull request information

- **Status**: Ready to merge
- **Kind of changes**: New env & example
- **Related PR or issue**: -

## Description

Multi-agent training environment for the scenario `HighwayRampsScenario` + an example to train on it.

Same states/actions/rewards as for `WaveAttenuationMergePOEnv` environment, adapted for multi-agent.


nanxintin:
Hi @AboudyKreidieh , I have an question about flow/envs/multiagent/highway.py. When we calculate the lead_head, I think we should use "self.k.vehicle.get_headway(rl_id)", not "self.k.vehicle.get_headway(lead_id)".

PR #677, opened 2019-08-07
Missing matplotlib in conda setup

cathywu:
## Pull request information

- **Status**: ready to merge
- **Kind of changes**: setup

## Description

Missing matplotlib in conda setup.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -17,6 +17,7 @@ dependencies:
         - nose2==0.8.0
         - cached_property
         - joblib==0.10.3
+        - matplotlib==3.0.0
         - dill
         - lz4
         - ray==0.6.1


AboudyKreidieh:
This is already in requirements.txt. Why do we need it here?

cathywu:
My understanding is that we should eventually be able to do the full install of flow without the `requirements.txt`. However, if conda fails at any point, all the remaining dependencies are not installed, and hence it's still useful to have a requirements file sitting around.

AboudyKreidieh:
if anything it should be the other way around. we need the `requirements.txt` file while running `pip install -e .` but not `environment.yml`. Also, anyone can always create their own environment and run the install there. Finally, people that don't want to use conda should have an option for installing flow that doesn't require it

eugenevinitsky:
Can't hurt to have it in both places.

PR #694, opened 2019-08-20
Evinitsky/setupfix

eugenevinitsky:
Remove stable baselines from environment.yml and requirements.txt. It requires an installation of MPI which causes peoples envs to break if they don't have it.


eugenevinitsky:
@AboudyKreidieh this is a two second review but super critical. The current setup instructions will break for anyone who doesn't have MPI installed.

kjang96:
LGTM

araffin:
Hello,
the next version of stable baselines (currently in the master branch) will support mpi less installation ;) (cf doc for details)

eugenevinitsky:
@araffin oh great! Thanks, once that's up on pip we'll change this so it can come installed.

Issue #718, opened 2019-09-11
Passing a Env instance to flow_params when using rllib


temetski:
Currently, flow only accepts strings when passing Env names to register in `gym`. A consequence of this is that anyone wishing to add new classes would have to modify the `__init__.py` files of `flow.envs`, and add new modules directly in the flow namespace. If instead we can pass class objects directly to `flow_params`, users can gain much more flexibility in extending the functionality of flow without tinkering with flow's source code. (see #192)

One possible way to attack this problem would be by modifying `make_create_env` directly. Any class has a `__module__` and `__name__` attribute, which can then be combined as 
```
entry_point = Class.__module__ + ':' + Class.__name__
```
This string can then be passed safely to `gym.envs.registration.register`.

eugenevinitsky:
Yes this is really good! Especially if we want users to be able to pip install flow this makes sense.

temetski:
I'll get a pull request ready. I'll start by overloading the `env_name` key first, so that we maintain compatibility with how Env names are currently passed, and see how you guys want to branch out from there.

eugenevinitsky:
Thanks! This is super helpful.

temetski:
Closed by #719

PR #719, opened 2019-09-12
Allows passing of Env and Scenario instance through flow_params

temetski:
## Pull request information

- **Status**: ready?
- **Related PR or issue**: resolves issue #718 

## Description

Allows for the passing of custom Environments as class instances rather than strings. Its current implementation does not break the old functionality, instead allowing for either the class instance or a string to be passed in `flow_params['env_name']`.

This comes with a caveat: the class instance has to be defined in a separate file (from the script which calls `make_create_env`) and imported in your main script.


temetski:
Current implementation allows for training a model, but fails when replaying the trained model via `visualizing_rllib.py`. Will investigate for now.

temetski:
Trained models can be visualized if we store the entire module path name (like `flow.envs.MergePOEnv`) in `flow_params['env_name']`. So in its current form, we can pass Env objects at train time, but this Env will still be string encoded in `flow_params`, so that the environment can be reloaded without importing the Env object.

The caveat is still that the Env object must be importable when `visualizer_rllib.py` is called. So either we run visualization in a folder that contains the Env file, or the Env file be installed as a python module (independently of flow).

temetski:
To make thing consistent, I propose that `flow_params` should store `env_name` and `network` as classes, while these classes get serialized through `FlowParamsEncoder` as `"module.ClassName"` strings. Decoding of the flow_params json would then reload these strings as classes.

PR #746, opened 2019-10-02
Fix bugs in utils/aimsun/load.py

umangs94:
## Pull request information

- **Status**: ready to merge

## Description

1. Updated the path to data.json
2. Added a check before querying centroid_config.origin_centroids and centroid_config.destination_centroids to prevent erroring out when the model doesn't have origin or destination centroids set.

### DIFF
--- a/flow/utils/aimsun/load.py
+++ b/flow/utils/aimsun/load.py
@@ -56,10 +56,16 @@ def get_dict_from_objects(sections, nodes, turnings, cen_connections):
     if not centroid_config:
         print('[load.py] ERROR: Centroid configuration ' +
               centroid_config_name + ' does not exist.')
-    for c in centroid_config.origin_centroids:
-        scenario_data['centroids'][c.id] = {'type': 'in'}
-    for c in centroid_config.destination_centroids:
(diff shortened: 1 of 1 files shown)


flow/utils/aimsun/load.py
```diff
@@ -132,7 +138,7 @@ def get_dict_from_objects(sections, nodes, turnings, cen_connections):
 model.load(template_path)
 
 # collect the simulation parameters
-params_file = 'flow/core/kernel/scenario/data.json'
+params_file = 'flow/core/kernel/network/data.json'
```
Yasharzf: In lines 177 and 186 we have similar bugs. Could you also fix them (replace 'scenario' with 'network')?

umangs94: In my local directory, those two files are still in `core/kernel/scenario`, not `network`. I don't think that they should be changed.

AboudyKreidieh: yup, please do

umangs94: Done!

AboudyKreidieh: great, thanks!

Issue #752, opened 2019-10-08
[Aimsun] Flow cannot be run outside flow directory when running Aimsun


temetski:
## Bug Description

Because Aimsun uses python2.7, flow is not visible as a module. When flow is run outside the flow directory, and a call to `flow/utils/aimsun/load.py` is made, it is unable to find `flow.utils.aimsun.scripting_api` and `flow.config` as flow is not installed as a python 2.7 module.

This would easily be fixed if Aimsun uses python 3.

## Bug Reproduce

File in which the bug occurred: `flow/utils/aimsun/load.py`

The exact command that you did run that caused the bug:
`python flow/examples/aimsun/grid.py`
(or any aimsun related script)

## Screenshots

Screenshots:
![image](https://user-images.githubusercontent.com/6328105/66418483-0db89b80-e9b7-11e9-8ac8-7f00c23eb5c0.png)

## Versions

- OS: macOS Mojave
- Flow version: 0.5.0.dev
- Flow commit number: 375fac9b

Yasharzf:
We need to switch to External Agent Interface of Aimsun which can help us avoid dealing with python 2.7 issues. But for now, you can export the path to PYTHONPATH (`export PYTHONPATH="/path/to/flow:$PYTHONPATH"`). We need to add to this to the installation docs.

temetski:
I'll work on a PR for this.

Yasharzf:
Great! Thanks!

temetski:
Addressed by #753.

PR #753, opened 2019-10-10
Aimsun fixes

temetski:
## Pull request information

- **Status**: ready to merge
- **Related PR or issue**: #752 

## Description

Mainly addresses issue #752 by setting `os.environ['PYTHONPATH'] = config.PROJECT_PATH`.
Also adds some minor bug fixes in loading osm templates (length was being set to sum of length of all lanes). Lastly, this PR gets rid of the annoying save prompt that pops up when running Aimsun.

### DIFF
--- a/flow/utils/aimsun/generate.py
+++ b/flow/utils/aimsun/generate.py
@@ -685,7 +685,7 @@ def create_meter(model, edge):
     """
     section = model.getCatalog().findByName(edge, model.getType("GKSection"))
     meter_length = 2
-    pos = section.getLanesLength2D() - meter_length
+    pos = section.length2D() - meter_length
     type = model.getType("GKMetering")
     cmd = model.createNewCmd(model.getType("GKSectionObject"))
     # TODO double check the zeros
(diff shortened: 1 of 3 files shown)


flow/utils/aimsun/generate.py
```diff
@@ -685,7 +685,7 @@ def create_meter(model, edge):
     """
     section = model.getCatalog().findByName(edge, model.getType("GKSection"))
     meter_length = 2
-    pos = section.getLanesLength2D() - meter_length
+    pos = section.length2D() - meter_length
```
AboudyKreidieh: is this a feature in a newer version of Aimsun? Why wasn't it broken before?

temetski: I think this was always there, but it's a method available to the parent class. Aimsun's documentation was difficult to navigate through. The reason this wasn't broken before is that this network only had single lanes per edge. `getLanesLength2D` returns the sum of the lengths of all lanes, which isn't appropriate for this use case.

PR #759, opened 2019-10-18
Example of how to reroute all vehicles

eugenevinitsky:
Adds an optional param that lets you reroute every vehicle if it is on its final edge.

### DIFF
--- a/flow/core/params.py
+++ b/flow/core/params.py
@@ -628,7 +628,8 @@ class EnvParams:
                  warmup_steps=0,
                  sims_per_step=1,
                  evaluate=False,
-                 clip_actions=True):
+                 clip_actions=True,
+                 reroute_if_final_edge=False):
         """Instantiate EnvParams."""
         self.additional_params = \
             additional_params if additional_params is not None else {}
@@ -637,6 +638,7 @@ class EnvParams:
         self.sims_per_step = sims_per_step
         self.evaluate = evaluate
         self.clip_actions = clip_actions
+        self.reroute_if_final_edge = reroute_if_final_edge
 
     def get_additional_param(self, key):
         """Return a variable from additional_params."""
--- a/flow/envs/base.py
+++ b/flow/envs/base.py
@@ -352,6 +352,10 @@ class Env(gym.Env):
 
             self.k.vehicle.choose_routes(routing_ids, routing_actions)
 
+            # reroute if on the final edge and request
+            if self.env_params.reroute_if_final_edge:
+                self.reroute_if_final_edge()
+
             self.apply_rl_actions(rl_actions)
 
             self.additional_command()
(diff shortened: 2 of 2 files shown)


nicofirst1:
The checking of the final edge is not valid for certain OSM maps (e.g. [this](https://gitlab.com/nicofirst1/mas_traffic/tree/master/FlowMas/maps/GroningenOSM)). For most vehicles the route is a single element list.

eugenevinitsky:
You can use self.k.vehicle.get_position(veh_id) to figure out the vehicle position and self.k.network.edge_length(edge) to get the edge length and figure out when you are near the edge. Adding that check for single edge routes should solve this issue.

nicofirst1:
Something like this?
```
            edge = self.k.vehicle.get_edge(veh_id)
            self.k.vehicle.get_position(veh_id)

            pos=self.k.vehicle.get_position(veh_id)
            length=self.k.network.edge_length(edge)

            eta=length*0.01


            # check if its on the final edge
            if edge == route[-1] and pos+eta>=length:
                type_id = self.k.vehicle.get_type(veh_id)
                # remove the vehicle
                self.k.vehicle.remove(veh_id)
```

nicofirst1:
Moreover I get this weird situation in which the previous if statement 
` if edge == route[-1] and pos+eta>=length` 
is _False_ but the vehicle id is in
 `sim_obs[tc.VAR_ARRIVED_VEHICLES_IDS]`  ([here](https://github.com/flow-project/flow/blob/master/flow/core/kernel/vehicle/traci.py#L130-L135)) 
therefore it is removed.

eugenevinitsky:
Yes that's the idea! One possible issue could be your value of eta is too small; if one of your edges is not too long and your speed is high you could travel off the edge before this condition catches it. Could you check if this is what's hapening?

A trick we used to use for this situation (that you could try) is to add long fake edges to every edge in the network that way there's zero chance of this happening. This is why there are weirdly long edges at the end of the grid example.

nicofirst1:
Using
`eta=length*0.1` 
Seems to solve the problem, I'll try with higher horizons (currently using 10).

Since this method seems a little forced in my opinion, I was wondering if there is a way to extract routes from an OSM map programmatically.

eugenevinitsky:
Cool! There isn't one that I know of, at least because it's an ill-posed problem. Technically any route is valid but you wouldn't take some routes (very long ones, routes that contain cycles). We have some work in progress to add this feature but it's not ready yet. If you want to contribute it it would be very appreciated!

nicofirst1:
I would love to contribute, but my project is due to next Sunday and I don't think I have the time.

eugenevinitsky:
Np. Hope it helps and let me know if anything comes up.

nicofirst1:
I found out that using the [add method](https://github.com/flow-project/flow/blob/master/flow/core/kernel/vehicle/traci.py#L1009), you do not reintroduce the agent in the [rl_ids](https://github.com/flow-project/flow/blob/master/flow/core/kernel/vehicle/traci.py#L39). This causes the samples to have inconsistent length, thus leading to [the error](https://github.com/ray-project/ray/issues/5913).

I tried using a custom add funtion:

```
    def add_rl(self, veh_id, type_id, edge, pos,sumo_obs):

        """See parent class."""
        if veh_id in self.master_kernel.network.rts:
            # If the vehicle has its own route, use that route. This is used in
            # the case of network templates.
            route_id = 'route{}_0'.format(veh_id)
        else:
            num_routes = len(self.master_kernel.network.rts[edge])
            frac = [val[1] for val in self.master_kernel.network.rts[edge]]
            route_id = 'route{}_{}'.format(edge, np.random.choice(
                [i for i in range(num_routes)], size=1, p=frac)[0])

        self.kernel_api.vehicle.addFull(
            veh_id,
            route_id,
            typeID=str(type_id),
            departLane="random",
            departPos=str(pos),
            departSpeed=str(0))

        self._add_departed(veh_id, type_id)



        # modify the number of vehicles and RL vehicles
        self.num_vehicles = len(self.get_ids())
        self.num_rl_vehicles = len(self.get_rl_ids())
```

But for some reason is not working as expected. Is there a way to reintroduce an RL vehicle immediately after the rerouting so that the next update will consider its observations?

eugenevinitsky:
There will be occasions where it’s not possible to reinsert the vehicle immediately. I’d just track the IDs and output a zero vector for that vehicle for the timestep it is missing. That shouldn’t be a problem as long as the amount of time it’s missing is small.

nicofirst1:
What are such occasions? Why wouldn't it be possible to reroute the agent immediately?

eugenevinitsky:
Where you want to insert them is blocked for example.

PR #761, opened 2019-10-23
Aimsun communication port no longer hardcoded

temetski:
## Pull request information

- **Status**: ready to merge

## Description

The TCP communication between Aimsun and Flow uses a hardcoded `PORT = 9999`. A consequence of this is that we are unable to use `rllib` for Aimsun scenarios. The process call calling `load.py` and `generate.py` can accept keyword arguments, which the two scripts store under author (a hack to fix a hack). The port number stored in this way can then be read by `run.py`.

Ports are generated by `sumolib.miscutils.getFreeSocketPort()`, which should pose no issues assuming the use-case where we only run Aimsun simulations.

### DIFF
--- a/flow/core/kernel/network/aimsun.py
+++ b/flow/core/kernel/network/aimsun.py
@@ -120,7 +120,7 @@ class AimsunKernelNetwork(BaseKernelNetwork):
                 f.write(template_path)
 
         # start the aimsun process
-        aimsun_call = [aimsun_path, "-script", script_path]
+        aimsun_call = [aimsun_path, "-script", script_path, str(self.sim_params.port)]
         self.aimsun_proc = subprocess.Popen(aimsun_call)
 
         # merge types into edges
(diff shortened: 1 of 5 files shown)


AboudyKreidieh:
Two question about this:

1. How is the new feature a hack? It seems reasonable to be passing the seed as an argument and storing it internally to the aimsun instance
2. What happens if port is set to None? This is the default value for `sim_params.port`

temetski:
@AboudyKreidieh It's a hack in the sense that it's not a proper fix, since I'm using the author field to pass down port info. But that shouldn't pose any issues. I have not tried `sim_params.port=None`, but currently `flow.envs.Env` sets 
```
        # FIXME: this is sumo-specific
        self.sim_params.port = sumolib.miscutils.getFreeSocketPort()
```
so the actual default behavior in master is not `sim_params.port=None`

AboudyKreidieh:
oh sweet, I was going to ask we add that if it wasn't there. In that case, LGTM

PR #765, opened 2019-10-28
Grid routing connections

kevin-thankyou-lin:
DO NOT MERGE


kevin-thankyou-lin:
DO NOT MERGE

kevin-thankyou-lin:
@eugenevinitsky the current code is here!

Non-RL works for traffic_light_grid and queue_grid. I stopped altering tl_grid halfway, because I made the queue_grid, so my changes to tl_grid code should be ignored.

PR #780, opened 2019-11-18
bump the high of traffic light grid po env

eugenevinitsky:
Change the observation space to handle the fact that values go slightly outside of it.


kjang96:
Once that's fixed, LGTM!

kjang96:
Actually I'm curious, what happens when there is no leader, so a vehicle's headway automatically returns min(1000, env.network.scenario_length..)? Your code doesn't actually change this logic, so I'm wondering about how this ever worked. Wouldn't you frequently have a leaderless vehicle returning a really high value?

kjang96:
Also caught one more spot that requires changing, in the update function in vehicle/traci.py around line 222 : self.__vehicles[veh_id]["headway"] = 1e+3
                self.__vehicles[veh_id]["follower_headway"] = 1e+3

eugenevinitsky:
Oh I missed these reviews! These are all good; thank you!

eugenevinitsky:
> Actually I'm curious, what happens when there is no leader, so a vehicle's headway automatically returns min(1000, env.network.scenario_length..)? Your code doesn't actually change this logic, so I'm wondering about how this ever worked. Wouldn't you frequently have a leaderless vehicle returning a really high value?

This is actually intentional. We want there to be a default high value indicating that there's no vehicle in front

kjang96:
> > Actually I'm curious, what happens when there is no leader, so a vehicle's headway automatically returns min(1000, env.network.scenario_length..)? Your code doesn't actually change this logic, so I'm wondering about how this ever worked. Wouldn't you frequently have a leaderless vehicle returning a really high value?
> 
> This is actually intentional. We want there to be a default high value indicating that there's no vehicle in front

Ah I see! So a `def has_leader(veh_id)` function could ostensibly look like `if tailway(veh_id) > [some_high_value]: return False` ? If it's a developer responsibility to modify their code such that they should be treating ultra-high returns of get_tailway or get_headway as "no leader" or "no tailway" maybe we should advertise that more. But maybe we do? I'm not sure

pengyuan-zhou:
works like a charm

Issue #777, opened 2019-11-11
Save movie at experiment end


eugenevinitsky:
It would be so, so convenient if post-training we immediately created and saved a movie of the trained policy. This would over the course of a long research project, save a ton of time. This could be done by running visuallizer_rllib but would require a way to extract the images.

PR #784, opened 2019-11-21
render to video using sumo-gui

temetski:
## Pull request information

- **Status**: in development
- **Related PR or issue**: #777 

Generates sumo screenshots using `self.k.kernel_api.gui.screenshot("View #0", 'screenshot.png')`.  These screenshots are saved to `~/flow_rendering` by default. These images should be convertible to a video using the `--save_render` argument in `visualizer_rllib.py`, or when passing `save_render=True` to SumoParams.

### DIFF
--- a/flow/envs/base.py
+++ b/flow/envs/base.py
@@ -7,6 +7,8 @@ import time
 import traceback
 import numpy as np
 import random
+import shutil
+import subprocess
 from flow.renderer.pyglet_renderer import PygletRenderer as Renderer
 from flow.utils.flow_warnings import deprecated_attribute
 
@@ -216,12 +218,8 @@ class Env(gym.Env):
         elif self.sim_params.render in [True, False]:
             # default to sumo-gui (if True) or sumo (if False)
             if (self.sim_params.render is True) and self.sim_params.save_render:
-                path = os.path.expanduser('~')+'/flow_rendering/'
-                dir_time = time.strftime("%Y-%m-%d-%H%M%S")
-                if not os.path.exists(path):
-                    os.mkdir(path)
-                self.path = path + dir_time
-                os.mkdir(self.path)
+                self.path = os.path.expanduser('~')+'/flow_rendering/' + self.network.name
+                os.makedirs(self.path, exist_ok=True)
         else:
             raise FatalFlowError(
                 'Mode %s is not supported!' % self.sim_params.render)
@@ -679,16 +677,13 @@ class Env(gym.Env):
                 self.renderer.close()
             # generate video
             elif (self.sim_params.render is True) and self.sim_params.save_render:
-                save_dir = os.path.expanduser('~') + '/flow_movies'
                 images_dir = self.path.split('/')[-1]
-                if not os.path.exists(save_dir):
-                    os.mkdir(save_dir)
-                speedup = 10  # speedup multiplier
+                speedup = 10  # multiplier: renders video so that `speedup` seconds is rendered in 1 real second
                 fps = speedup//self.sim_step
-                os_cmd = "ffmpeg -y -r {fps} -i {path}/frame_%06d.png".format(path=self.path, fps=fps)
-                os_cmd += " -pix_fmt yuv420p " + self.path+"/%s.mp4" % images_dir
-                os_cmd += "&& cp " + self.path+"/%s.mp4 " % images_dir + save_dir + "/"
-                os.system(os_cmd)  # only works in Unix
+                p = subprocess.Popen(["ffmpeg", "-y", "-r", str(fps), "-i", self.path+"/frame_%06d.png",
+                                      "-pix_fmt", "yuv420p", "%s/../%s.mp4" % (self.path, images_dir)])
+                p.wait()
+                shutil.rmtree(self.path)
         except FileNotFoundError:
             # Skip automatic termination. Connection is probably already closed
             print(traceback.format_exc())
--- a/flow/envs/base.py
+++ b/flow/envs/base.py
@@ -214,7 +214,14 @@ class Env(gym.Env):
             # render a frame
             self.render(reset=True)
         elif self.sim_params.render in [True, False]:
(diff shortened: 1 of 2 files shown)


eugenevinitsky:
I'll review this tonight, thanks so much for this! This should let us auto-make videos and sync them to s3 when the experiments are done. It's a HUGE timesaver.

temetski:
@eugenevinitsky I actually am not sure if sumo-gui plays well with headless displays. Hopefully it does.

I didn't add it to the ffmpeg call in `visualizer_rllib`, but you can pass a framerate that speeds up playback by some factor (for long videos that are too boring to watch in slow motion, when sim_step is small).

flow/envs/base.py
```diff
@@ -214,7 +214,14 @@ def __init__(self,
             # render a frame
             self.render(reset=True)
         elif self.sim_params.render in [True, False]:
-            pass  # default to sumo-gui (if True) or sumo (if False)
+            # default to sumo-gui (if True) or sumo (if False)
+            if (self.sim_params.render is True) and self.sim_params.save_render:
+                path = os.path.expanduser('~')+'/flow_rendering/'
```
eugenevinitsky: Why is there both a flow_rendering path and a flow_movies path?

flow/envs/base.py
```diff
@@ -214,7 +214,14 @@ def __init__(self,
...
+                dir_time = time.strftime("%Y-%m-%d-%H%M%S")
+                if not os.path.exists(path):
+                    os.mkdir(path)
+                self.path = path + dir_time
```
eugenevinitsky: Think it might be worthwhile to add the exp_title to the path so that it's a little easier to track what movie you want to look at

flow/envs/base.py
```diff
@@ -670,6 +677,18 @@ def terminate(self):
...
+                save_dir = os.path.expanduser('~') + '/flow_movies'
+                images_dir = self.path.split('/')[-1]
+                if not os.path.exists(save_dir):
+                    os.mkdir(save_dir)
+                speedup = 10  # speedup multiplier
+                fps = speedup//self.sim_step
+                os_cmd = "ffmpeg -y -r {fps} -i {path}/frame_%06d.png".format(path=self.path, fps=fps)
+                os_cmd += " -pix_fmt yuv420p " + self.path+"/%s.mp4" % images_dir
+                os_cmd += "&& cp " + self.path+"/%s.mp4 " % images_dir + save_dir + "/"
+                os.system(os_cmd)  # only works in Unix
```
eugenevinitsky: You can subprocess.Popen for a command that works in both unix and mac

flow/envs/base.py
```diff
@@ -670,6 +677,18 @@ def terminate(self):
             # close pyglet renderer
             if self.sim_params.render in ['gray', 'dgray', 'rgb', 'drgb']:
                 self.renderer.close()
+            # generate video
+            elif (self.sim_params.render is True) and self.sim_params.save_render:
+                save_dir = os.path.expanduser('~') + '/flow_movies'
+                images_dir = self.path.split('/')[-1]
+                if not os.path.exists(save_dir):
```
eugenevinitsky: As mentioned above, slightly confused about this second path

flow/envs/base.py
```diff
@@ -670,6 +677,18 @@ def terminate(self):
...
+                    os.mkdir(save_dir)
+                speedup = 10  # speedup multiplier
+                fps = speedup//self.sim_step
```
eugenevinitsky: could you leave a comment explaining what this is? I can't quite figure it out from these lines alone.

flow/envs/base.py
temetski: This was patterned after what already exists in `visualizer_rlllib.py`.

flow/envs/base.py
temetski: It adjusts the framerate so that you're watching `speedup` simulation seconds per 1 realtime second. I'll make the comment more explicit.

flow/envs/base.py
temetski: I actually didn't want to follow this pattern that I saw in `visuallizer_rllib.py` (i was assuming that's how you guys were using this). If we merge flow_movies with flow_rendering, subprocess.Popen can be used. (subprocess somehow cannot be configured to wait for ffmpeg to finish running, hence chaining a copy command wouldn't work.)

flow/envs/base.py
temetski: Good catch. Is the experiment path visible to an environment?

eugenevinitsky: Hmm, it appears it is not but it's worth adding. If we don't get it in in this PR could you add it as an issue?

flow/envs/base.py
eugenevinitsky: https://stackoverflow.com/questions/15107714/wait-process-until-all-subprocess-finish

flow/envs/base.py
eugenevinitsky: I see. I think we should probably merge this into one folder instead of there being two. It seems like that would reduce complexity rather than raise it.

flow/envs/base.py
eugenevinitsky: Last thing; can we clean and remove the screenshots after this command completes?

temetski: Yea, i did check out wait, but it doesn't work for ffmpeg for some strange reason. Yes, we can clean up all auxiliary files. I'm adding that in.

flow/envs/base.py
temetski: Looks like experiment titles are being passed as network names, so I was able to work with that.

flow/envs/base.py
temetski: Done. Sticking with `flow_rendering` for consistency with pyglet functionality.

eugenevinitsky (approved):
LGTM! Thanks for the PR Damian

eugenevinitsky:
Feel free to merge it whenever you're comfortable @temetski

PR #817, opened 2020-01-17
Added Pedestrian Code for New Traffic_Light_Grid

NickL77:


examples/sumo/traffic_light_grid.py
```diff
@@ -384,7 +438,12 @@ def generate_tl_phases(phase_type, horiz_lanes, vert_lanes):
 
 
 if __name__ == "__main__":
+
+    pedestrians = False
+    if '--pedestrians' in sys.argv:
+        pedestrians = True
```
eugenevinitsky: Can you use argparse for this? Just a preference since we use it elsewhere.

flow/core/params.py
```diff
@@ -349,6 +349,29 @@ def get_type(self, veh_id):
...
+    def __init__(self):
+
+        self.ids = []
+        self.num_pedestrians = 0
+        self.params = {}
+
+        self.__pedestrians = collections.OrderedDict()
+
+    def add(self, ped_id, depart_time, start, end):
+        self.ids.append(ped_id)
```
eugenevinitsky: This needs a docstring explaining what start and end are

scripts/setup_aimsun.sh
```diff
@@ -5,7 +5,8 @@ conda config --append channels https://repo.anaconda.com/pkgs/free
 conda config --append channels https://repo.anaconda.com/pkgs/pro
 
 # create the conda environment
-conda create -y -n aimsun_flow python=2.7.4
+#conda create -y -n aimsun_flow python=2.7.4
+conda create -y -n aimsun_flow python=2.7.13
```
eugenevinitsky: Why is this here?

PR #834, opened 2020-02-18
Fix Vehicle Vision and Add DocStrings

NickL77:
Fix vehicle blocking itself out of view, and added docstrings to util functions.

### DIFF
--- a/flow/core/kernel/util.py
+++ b/flow/core/kernel/util.py
@@ -27,14 +27,10 @@ def observed(position, orientation, target_position, fov=90, looking_distance=50
     # edge case where both objects are at the same position
     if delta_x == 0 and delta_y == 0:
         return True
-    
+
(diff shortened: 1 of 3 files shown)


flow/core/kernel/util.py
```diff
@@ -1,16 +1,16 @@
...
 def observed(position, orientation, target_position, fov=90, looking_distance=50):
-
-    """ Checks if a single vehicle/pedestrian can see another vehicle/pedestrian
+    """Check if a single vehicle/pedestrian can see another vehicle/pedestrian.
 
     Parameters
-    ---------------------
-    position : 2D tuple
+    ----------
+    position : tuple of (double, double)
```
eugenevinitsky: nit: these are just floats in python. floats by default have 64 bit precision

PR #836, opened 2020-02-20
Fix the extra ray windows that pop up

eugenevinitsky:
This removes the extraneous windows that Ray opens.

### DIFF
--- a/flow/envs/base.py
+++ b/flow/envs/base.py
@@ -129,7 +129,10 @@ class Env(gym.Env):
         self.network = scenario if scenario is not None else network
         self.net_params = self.network.net_params
         self.initial_config = self.network.initial_config
-        self.sim_params = sim_params
+        self.sim_params = deepcopy(sim_params)
+        # check whether we should be rendering
+        self.should_render = self.sim_params.render
+        self.sim_params.render = False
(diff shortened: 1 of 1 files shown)


eugenevinitsky:
Credit to @kevin-thankyou-lin for actually figuring out why this was happening.

flow/envs/base.py
```diff
@@ -426,6 +429,13 @@ def reset(self):
         # reset the time counter
         self.time_counter = 0
 
+        # Now that we've passed the possibly fake init steps some rl libraries
+        # do, we can feel free to actually render things
+        if self.should_render:
+            self.sim_params.render = True
+            # got to restart the simulation to make it actually display anything
+            self.restart_simulation(self.sim_params)
```
kevin-thankyou-lin: @eugenevinitsky line 437 actually creates an extra restart_simulation (at least when I'm running bayesian_1_env.py)

Line 444 creates another restart instance each time:

self.restart_simulation(self.sim_params)

And s, in the forked repo, I just deleted the one self.restart_simulation(self.sim_params) line. Not sure if we want to do it here?

eugenevinitsky: Great catch

PR #846, opened 2020-03-02
Add a nonlocal example of follower stopper as requested

eugenevinitsky:


flow/controllers/velocity_controllers.py
```diff
@@ -116,6 +116,53 @@ def get_accel(self, env):
...
 
 
+class NonLocalFollowerStopper(FollowerStopper):
+    """Follower stopper that uses the average system speed to compute its acceleration."""
+
+    def get_accel(self, env):
+        """See parent class."""
+        lead_id = env.k.vehicle.get_leader(self.veh_id)
+        this_vel = env.k.vehicle.get_speed(self.veh_id)
+        lead_vel = np.mean(env.k.vehicle.get_speed(env.k.vehicle.get_ids()))
```
AboudyKreidieh: My understanding was they wanted us to do something like

```
self.v_des = np.mean(env.k.vehicle.get_speed(env.k.vehicle.get_ids()))
```

Are you sure this is what they wanted?

eugenevinitsky: Honestly not clear on what they wanted but what you have above makes more sense.

Yasharzf: @AboudyKreidieh is right. They wanted to assign the speed to the average speed.

AboudyKreidieh: k, if we all agree I can modify and merge it

eugenevinitsky: Fixed.

PR #848, opened 2020-03-04
Color vehicles named AV red

eugenevinitsky:
Any vehicle that contains the substring av will be colored red. This is an unknown unknown but I'm having trouble thinking of a cleaner solution.


eugenevinitsky:
Superseded by Kanaad's PR

PR #849, opened 2020-03-04
Add flag auto_color flag to SumoParams and disable coloring if flag is off.

kanaadp:
## Pull request information

- **Status**: ? Ready to Merge
- **Kind of changes**: ? New Feature
- **Related PR or issue**: ? N/A

## Description

Added check in traci_kernel to turn off auto_coloring (for sumo). This allows us to specify custom colors for vehicle types, inflows, and routes. 

i.e. 
Screen Shot 2020-03-04 at 12 19 16 PM

Colors can be set via:
Screen Shot 2020-03-04 at 12 19 28 PM

or 

Screen Shot 2020-03-04 at 12 19 45 PM

### DIFF
--- a/flow/core/kernel/vehicle/traci.py
+++ b/flow/core/kernel/vehicle/traci.py
@@ -987,7 +987,7 @@ class TraCIVehicle(KernelVehicle):
 
         for veh_id in self.get_rl_ids():
             try:
-                if force_update or not 'color' in self.type_parameters[self.get_type(veh_id)]:
+                if force_update or 'color' not in self.type_parameters[self.get_type(veh_id)]:
                     # color rl vehicles red
                     self.set_color(veh_id=veh_id, color=RED)
             except (FatalTraCIError, TraCIException) as e:
@@ -997,7 +997,7 @@ class TraCIVehicle(KernelVehicle):
         for veh_id in self.get_human_ids():
             try:
                 color = CYAN if veh_id in self.get_observed_ids() else WHITE
-                if force_update or not 'color' in self.type_parameters[self.get_type(veh_id)]:
+                if force_update or 'color' not in self.type_parameters[self.get_type(veh_id)]:
                     self.set_color(veh_id=veh_id, color=color)
             except (FatalTraCIError, TraCIException) as e:
                 print('Error when updating human vehicle colors:', e)
--- a/flow/core/kernel/vehicle/traci.py
+++ b/flow/core/kernel/vehicle/traci.py
@@ -71,10 +71,7 @@ class TraCIVehicle(KernelVehicle):
         self._arrived_rl_ids = []
 
         # whether or not to automatically color vehicles
-        try:
-            self._color_vehicles = sim_params.color_vehicles
-        except AttributeError:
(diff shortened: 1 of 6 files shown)


eugenevinitsky:
It seems like auto_color_vehicles is unnecessary and likely to cause confusion. Why not just leave default coloring on and override it if a color is specified?

eugenevinitsky:
Otherwise, great minds think alike: #848

flow/core/params.py
```diff
@@ -281,6 +282,9 @@ def add(self,
         type_params.update(car_following_params.controller_params)
         type_params.update(lane_change_params.controller_params)
 
+        if color:
+            type_params['color'] = color
```
eugenevinitsky: This works?! That's amazing!

AboudyKreidieh:
We're gonna end up in a situation here where one feature (`color`) will not work unless another default parameter is changed (in this case `auto_color_vehicles`). Should we default `auto_color_vehicles` to False, stick with the earlier version of this PR, or are we okay with this?

eugenevinitsky:
This is an improvement on my PR, but, I think we can just remove auto_color_vehicles completely and just override it if and only if a color is provided.

kanaadp:
you'd have to check if any inflow, vehicle, or route has a color defined before removing

eugenevinitsky:
Or you can just have color default to None and if it's None you don't override.

eugenevinitsky:
Some mechanism is needed, expecting people to remember to set auto_color_vehicles to false when you add colors will cause a lot of confusion,.

flow/core/kernel/vehicle/traci.py
```diff
@@ -974,26 +974,31 @@ def get_x_by_id(self, veh_id):
...
         - cyan: observed human-driven vehicles
         """
+        if not self._color_vehicles:
+            return
+
         for veh_id in self.get_rl_ids():
             try:
-                # color rl vehicles red
-                self.set_color(veh_id=veh_id, color=RED)
+                if force_update or 'color' not in self.type_parameters[self.get_type(veh_id)]:
```
eugenevinitsky: Please add one comment explaining this line. It's the one thing that's confusing

flow/core/kernel/vehicle/traci.py
```diff
@@ -974,26 +974,31 @@ def get_x_by_id(self, veh_id):
...
+                    self.set_color(veh_id=veh_id, color=RED)
             except (FatalTraCIError, TraCIException) as e:
                 print('Error when updating rl vehicle colors:', e)
 
         # color vehicles white if not observed and cyan if observed
         for veh_id in self.get_human_ids():
             try:
                 color = CYAN if veh_id in self.get_observed_ids() else WHITE
-                self.set_color(veh_id=veh_id, color=color)
+                if force_update or 'color' not in self.type_parameters[self.get_type(veh_id)]:
```
eugenevinitsky: Ditto here.

kanaadp:
@eugenevinitsky and @AboudyKreidieh made it a lil cleaner, lmk

PR #894, opened 2020-03-30
Noise fix

eugenevinitsky:
Noise is scaled by sqrt of sim step as suggested by theory group


eugenevinitsky:
Thanks for the review!
1. We're not requiring this branch to pass tests outside of Flake8 and pydoc. That's the point.
2. The noise is actually correct, that's what Benni wanted and what sets off waves.

PR #898, opened 2020-04-05
add traffic scale option to running with sumo

zpymyyn:
as the title

### DIFF
--- a/flow/core/params.py
+++ b/flow/core/params.py
@@ -395,6 +395,7 @@ class SimParams(object):
 
     def __init__(self,
                  sim_step=0.1,
+                 scale=1,
                  render=False,
                  restart_instance=False,
                  emission_path=None,
@@ -405,6 +406,7 @@ class SimParams(object):
                  force_color_update=False):
         """Instantiate SimParams."""
         self.sim_step = sim_step
+        self.scale = scale
         self.render = render
         self.restart_instance = restart_instance
         self.emission_path = emission_path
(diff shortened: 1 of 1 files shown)


flow/core/params.py
```diff
@@ -395,6 +395,7 @@ class SimParams(object):
 
     def __init__(self,
                  sim_step=0.1,
+                 scale=1.0,
```
eugenevinitsky, line 398: Hi, could you add an explanation of what this does to the docstring?

flow/core/params.py
```diff
@@ -573,6 +575,7 @@ class SumoParams(SimParams):
     def __init__(self,
                  port=None,
                  sim_step=0.1,
+                 scale=1.0,
```
eugenevinitsky, line 578: Same as above.

flow/core/params.py
zpymyyn: Sure thing, done :)

flow/core/params.py
zpymyyn: done:)

zpymyyn:
@eugenevinitsky Hi, is there any issue with this snippet?

pengyuan-zhou:
> Where is scale actually passed to SUMO?

Hi, it's been a while and I can't recall. But it's an official option of sumo as shown [here](https://sumo.dlr.de/docs/SUMO-GUI.html#scaling_the_amount_of_traffic)

PR #907, opened 2020-04-14
Aimsun bugs

Yasharzf:
## Pull request information

- **Status**: ready to merge 
- **Related PR or issue**: 

## Description

Fixed aimsun template file
Fixed a bug in aimsun network close function. 
Fixed a bug in simulate.py. When flag.aimsun is true, we need to pass AimsunParams instead of SumoParams


examples/simulate.py
```diff
@@ -6,6 +6,8 @@
 import argparse
 import sys
 from flow.core.experiment import Experiment
+from flow.core.params import AimsunParams
+
```
AboudyKreidieh, line 10:
```suggestion
from flow.core.params import AimsunParams
```

PR #909, opened 2020-04-15
Evinitsky/straight road pr

eugenevinitsky:
add a straight road example for faster training
    swap out the reward function with one that actual changes as a function of speed


flow/core/params.py
```diff
@@ -637,6 +637,8 @@ class EnvParams:
         specifies whether to clip actions from the policy by their range when
         they are inputted to the reward function. Note that the actions are
         still clipped before they are provided to `apply_rl_actions`.
+    done_at_exit : bool, optional
+        If true, done is returned as True when the vehicle exits.
```
AboudyKreidieh, line 641:
```suggestion
        If true, done is returned as True when the vehicle exits. This is only 
        applied to multi-agent environments.
```

eugenevinitsky: Good catch.

AboudyKreidieh:
Finished first round of reviews. Let me know when things are updated and I'll look at it again.

eugenevinitsky:
Am I missing some comments? Seems like there's just the one comment.

AboudyKreidieh:
woops, needed to submit them. they're up now

eugenevinitsky:
Thanks for the review! Comments have been addressed.

PR #918, opened 2020-04-30
Add option to reroute exiting vehicles back into the network

eugenevinitsky:
If you set reroute_on_exit to True, any vehicle that didn't exist during warmup_steps will be removed, and any vehicle that exists at the end of warmup_steps will be placed back into the network when it exits.


flow/envs/multiagent/base.py
```diff
@@ -103,6 +107,7 @@ def step(self, rl_actions):
 
             # stop collecting new simulation steps if there is a collision
             if crash:
+                print('A CRASH! A CRASH!!!!!! AAAAAAAAAH!!!!!')
```
kjang96, line 110: NOOOOOOOO

flow/envs/base.py
```diff
@@ -148,6 +148,10 @@ def __init__(self,
         self.state = None
         self.obs_var_labels = []
 
+        # track IDs that have ever been observed in the system
```
kjang96, line 151: Can you also incorporate the changes you made in multiagent/base.py into the singleagent env for symmetry? More specifically, add logic that updates `self.observed_ids`

flow/envs/multiagent/i210.py
```diff
@@ -156,53 +159,67 @@ def compute_reward(self, rl_actions, **kwargs):
...
         return rewards
 
     def additional_command(self):
         """See parent class.
 
-        Define which vehicles are observed for visualization purposes.
+        Define which vehicles are observed for visualization purposes. Additionally, optionally reroute vehicles
+        back once they have exited.
         """
+        super().additional_command()
```
kjang96, line 180: Why is this necessary? If I"m not mistaken, super's additional command is defined to be an abstract method (the body is just 'pass')

flow/envs/multiagent/i210.py
```diff
@@ -156,53 +159,67 @@ def compute_reward(self, rl_actions, **kwargs):
...
+                    # TODO(@evinitsky) select the lane and speed a bit more cleanly
+                    # Note, the position is 10 so you are not overlapping with the inflow car that is being removed.
+                    # this allows the vehicle to be immediately inserted.
+                    self.k.vehicle.add(
+                        veh_id=veh_id,
+                        edge=self.entrance_edge,
+                        type_id=str(type_id),
+                        lane=str(lane),
+                        pos="10.0",
+                        speed="23.0")
```
kjang96, line 215: Should this be desired_velocity or something?

flow/envs/multiagent/i210.py
```diff
@@ -243,13 +260,26 @@ def veh_statistics(self, rl_id):
         lane = (self.k.vehicle.get_lane(rl_id) + 1) / 10.0
         return np.array([speed, lane])
```
kjang96, line 262: good catch!

examples/exp_configs/rl/singleagent/singleagent_straightroad.py
```diff
@@ -0,0 +1,163 @@
+"""Multi-agent highway with ramps example.
+Trains a non-constant number of agents, all sharing the same policy, on the
```
kjang96, line 2: Can you make it so the expconfigs for SA and MA are both either {ma/sa}_straight_road.py or {ma/sa}_straightroad.py ?

kjang96:
Also what's the purpose of the observed_ids set?

flow/envs/base.py
eugenevinitsky: Sure thing.

flow/envs/multiagent/i210.py
eugenevinitsky: It's just good practice to call the super() method, no?

flow/envs/multiagent/i210.py
eugenevinitsky: This is problematic but it just needs to match the inflow conditions.

examples/exp_configs/rl/singleagent/singleagent_straightroad.py
eugenevinitsky: yeah good point

eugenevinitsky:
Observed ids are used to track which vehicles I should actually be putting back into the system and to throw out vehicles that aren't those vehicles

PR #926, opened 2020-05-09
changed _departed_ids, and _arrived_ids in the update function

Yasharzf:
## Pull request information

- **Status**:  ready to merge
- **Related PR or issue**: 

## Description

In vehicle/traci, we are storing all departed and arrived vehicle IDs for each simulation step. Every simulation step we are appending new vehicle list to _departed_ids and _arrived_ids. In large networks or high horizon times, this can be expensive and results in significant slow down. I changed _departed_ids and _arrived_ids to just store the current ids. I didn't totally delete them since we may need them (e.g., as in MultiEnv)

### DIFF
--- a/flow/core/kernel/vehicle/traci.py
+++ b/flow/core/kernel/vehicle/traci.py
@@ -211,11 +211,10 @@ class TraCIVehicle(KernelVehicle):
                     self.__vehicles[veh_id]["last_lc"] = self.time_counter
 
             # updated the list of departed and arrived vehicles
-            self._num_departed.append(
-                len(sim_obs[tc.VAR_DEPARTED_VEHICLES_IDS]))
-            self._num_arrived.append(len(sim_obs[tc.VAR_ARRIVED_VEHICLES_IDS]))
-            self._departed_ids.append(sim_obs[tc.VAR_DEPARTED_VEHICLES_IDS])
-            self._arrived_ids.append(sim_obs[tc.VAR_ARRIVED_VEHICLES_IDS])
+            self._num_departed.append(sim_obs[tc.VAR_LOADED_VEHICLES_NUMBER])
+            self._num_arrived.append(sim_obs[tc.VAR_ARRIVED_VEHICLES_NUMBER])
+            self._departed_ids = sim_obs[tc.VAR_DEPARTED_VEHICLES_IDS]
+            self._arrived_ids = sim_obs[tc.VAR_ARRIVED_VEHICLES_IDS]
 
             # update the number of not departed vehicles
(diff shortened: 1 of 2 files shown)


flow/core/kernel/vehicle/traci.py
```diff
@@ -518,7 +517,7 @@ def get_num_arrived(self):
     def get_arrived_ids(self):
         """See parent class."""
         if len(self._arrived_ids) > 0:
-            return self._arrived_ids[-1]
+            return self._arrived_ids
```
eugenevinitsky, line 520: why this change? this affects a lot of the codebase. Otherwise LGTM

flow/core/kernel/vehicle/traci.py
```diff
@@ -532,7 +531,7 @@ def get_arrived_rl_ids(self):
     def get_departed_ids(self):
         """See parent class."""
         if len(self._departed_ids) > 0:
-            return self._departed_ids[-1]
+            return self._departed_ids
```
eugenevinitsky, line 534: same here

flow/core/kernel/vehicle/traci.py
Yasharzf: self._arrived_ids was a list storing ids for each step. in this function we just want the ids in the last step ([-1]). Since I don't have a list anymore, I deleted [-1].

eugenevinitsky: So if it's not a list anymore why does it have a len method?

Yasharzf: good catch!

flow/core/kernel/vehicle/traci.py
```diff
@@ -211,11 +211,10 @@ def update(self, reset):
                     self.__vehicles[veh_id]["last_lc"] = self.time_counter
 
             # updated the list of departed and arrived vehicles
-            self._num_departed.append(
-                len(sim_obs[tc.VAR_DEPARTED_VEHICLES_IDS]))
-            self._num_arrived.append(len(sim_obs[tc.VAR_ARRIVED_VEHICLES_IDS]))
-            self._departed_ids.append(sim_obs[tc.VAR_DEPARTED_VEHICLES_IDS])
-            self._arrived_ids.append(sim_obs[tc.VAR_ARRIVED_VEHICLES_IDS])
+            self._num_departed.append(sim_obs[tc.VAR_LOADED_VEHICLES_NUMBER])
+            self._num_arrived.append(sim_obs[tc.VAR_ARRIVED_VEHICLES_NUMBER])
```
eugenevinitsky, line 215: as long as there is a getter for these, this LGTM

PR #933, opened 2020-05-13
Mpg reward2

eugenevinitsky:
MPG reward and MP-joule reward. You can train directly on these.


eugenevinitsky:
This also contains the merge from master that brings aboudy's changes in

PR #939, opened 2020-05-21
Add several accelerations to custom output

liljonnystyle:
## Pull request information

- **Status**: ready to merge

## Description

Adding more accelerations to the custom data pipeline output.
- Target acceleration without noise and without failsafes (`target_accel_no_noise_no_failsafe`, new)
- Target acceleration without noise and with failsafes (`target_accel_no_noise_with_failsafe`, formerly `accel_without_noise`)
- Target acceleration with noise and without failsafes (`target_accel_with_noise_no_failsafe`, new)
- Target acceleration with noise and with failsafes (`target_accel_with_noise_with_failsafe`, formerly `acceleration`)
- Realized acceleration (no change)


flow/core/kernel/vehicle/traci.py
```diff
@@ -113,7 +113,10 @@ def initialize(self, vehicles):
                 self.__vehicles[veh_id] = dict()
                 self.__vehicles[veh_id]['type'] = typ['veh_id']
                 self.__vehicles[veh_id]['initial_speed'] = typ['initial_speed']
-                self.__vehicles[veh_id]["accel_without_noise"] = None
+                self.__vehicles[veh_id]["accel_no_noise_no_failsafe"] = None
+                self.__vehicles[veh_id]["accel_no_noise_with_failsafe"] = None
+                self.__vehicles[veh_id]["accel_with_noise_no_failsafe"] = None
+                self.__vehicles[veh_id]["accel_with_noise_with_failsafe"] = None
```
brentgryffindor, line 119: Since there is a key existence check in all the get methods below. There is no need to initialize them here.

flow/core/kernel/vehicle/traci.py
```diff
@@ -964,6 +967,20 @@ def apply_acceleration(self, veh_ids, acc):
                 next_vel = max([this_vel + acc[i] * self.sim_step, 0])
                 self.kernel_api.vehicle.slowDown(vid, next_vel, 1e-3)
 
+    def apply_acceleration_not_smooth(self, veh_ids, acc):
+        """See parent class."""
+        # to hand the case of a single vehicle
```
eugenevinitsky, line 972: nit: to handle

flow/core/kernel/vehicle/traci.py
```diff
@@ -964,6 +967,20 @@ def apply_acceleration(self, veh_ids, acc):
...
+        if type(veh_ids) == str:
+            veh_ids = [veh_ids]
+            acc = [acc]
+
+        for i, vid in enumerate(veh_ids):
+            if acc[i] is not None and vid in self.get_ids():
+                self.__vehicles[vid]["accel"] = acc[i]
+                this_vel = self.get_speed(vid)
+                next_vel = max([this_vel + acc[i] * self.sim_step, 0])
+                self.kernel_api.vehicle.setSpeed(vid, next_vel)
```
eugenevinitsky, line 982: how is this different than the normal method?

Yasharzf: the normal method uses slowDown() method which applies some sort of smoothing. here setSpeed() method is used which doesn't apply smoothing.

PR #943, opened 2020-05-21
Control range added, affects reward and state

kjang96:
## Pull request information

- **Status**: in development
- **Kind of changes**: new feach (bug fix / new feature / documentation...)

## Description

Added control range. You can sit this via additional env params. Control range only affects get_state and reward. Logic is pretty much just: if veh_id not in the control range, then don't include its info in get_state or compute_reward


eugenevinitsky:
For full comparison we also need a way to apply control_range to follower_stopper so that when we compare to it it's running the same control structure. Otherwise it's not apples to apples

kjang96:
We can keep this PR if we want to add the ability to express control range in terms of meters rather than edges.

PR #944, opened 2020-05-22
Energy class for inventorying multiple energy models

liljonnystyle:
## Pull request information

- **Status**: ready to merge

## Description

This is a new implementation of a base energy class along with a few child classes to help us inventory the multiple energy models we plan to have.

@liljonnystyle is uploading on behalf of @joyncarpio to initiate tracking of code changes.


flow/core/kernel/energy.py
```diff
@@ -0,0 +1,119 @@
...
+
+class ToyotaPriusEVModel(EnergyModel):
+
+    def __init__(self,kernel):
+        self.k = kernel
+
+        # download file from s3 bucket
+        s3 = boto3.client('s3')
+        s3.download_file('toyota.restricted','prius_test.pkl','PriusEVModel.pkl') #move to init
+        with open('PriusEVModel.pkl','rb')as file:
```
liljonnystyle, line 70: missing a couple spaces after commas, and ')'

and you can remove the comment

flow/core/kernel/energy.py
```diff
@@ -0,0 +1,119 @@
...
+            prius_energy = pickle.load(file) #self.prius_energy
+        # delete pickle file
```
liljonnystyle, line 72: yes to both comments: add 'self.' and then be sure to delete it afterwards.

flow/core/kernel/energy.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            old_speed = speed
+
+        accel = (speed - old_speed)/0.1
+        power = M * speed * accel + M * g * Cr * speed + 0.5 * rho * A * Ca * speed ** 3
+
+        return power
+
+
+class ToyotaPriusEVModel(EnergyModel):
```
liljonnystyle, line 62: I think we can define a `ToyotaModel` class here, and recycle the s3/pickle code for both Prius and Tacoma and future models

flow/core/kernel/energy.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def __init__(self,kernel):
+        self.vehicle = vehicle
+
+    def get_energy(self, veh_id):
+        #return self.calc_energy(veh_id)
+        pass
+
+
+class PowerDemandModel(EnergyModel):
```
liljonnystyle, line 23: Not necessary, but I think we could move this and other child classes to other file(s). Maybe vehicle power demand in its own file, and then all Toyota models in one file?

also remember to implement an "EV" version and a "Gasoline" version of the vehicle power demand model

flow/core/kernel/energy.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def get_energy(self, veh_id)
+        # if we know the constants for other vehicle types, we may put them here as well
+        speed = self.k.get_speed(veh_id)
+        if veh_id in self.k.old_speeds:
+            old_speed = self.k.old_speeds[veh_id]
+        else:
+            old_speed = speed
+
+        accel = (speed - old_speed)/0.1
```
liljonnystyle, line 56: divide by the sim_step here (you need to check, does it come from the env?)

flow/core/kernel/energy.py
```diff
@@ -0,0 +1,119 @@
...
+
+        old_soc = self.k.get_soc(veh_id)
+        grade = 0 
+
+        socdot = prius_energy(old_soc,speed,accel,grade)
```
liljonnystyle, line 88: socdot is equivalent to power, so maybe this is what we should output from this function -- although we need to do some unit conversion from a percentage of battery power to watts

flow/core/kernel/vehicle/traci.py
```diff
@@ -113,6 +124,13 @@ def initialize(self, vehicles):
                 self.__vehicles[veh_id] = dict()
                 self.__vehicles[veh_id]['type'] = typ['veh_id']
                 self.__vehicles[veh_id]['initial_speed'] = typ['initial_speed']
+                # self.energy.initialize(veh_id)
+                # energy.intialize(veh_id)
+                self.__vehicles[veh_id]['SOC'] = float(random.randrange(8, 92))/100 #move to experiment code
+                # # Assume level of fuel is 10-100% of capacity
+                # # Tacoma has 21.1 gallon capacity = 79.87219 L
+                # # Use density of gasoline = 748.9 g/L (needs to convert b/c fc is in g/s)
+                self.__vehicles[veh_id]['fuel'] = float(random.randrange(598163, 5981628))/100
```
liljonnystyle, line 133: i think we may want to avoid using anything random here because everything else here is deterministic -- but i'm not sure yet what's a better alternative.

flow/core/kernel/vehicle/traci.py
```diff
@@ -88,6 +93,12 @@ def __init__(self,
         # old speeds used to compute accelerations
         self.previous_speeds = {}
 
+        self.energy = Energy(self)
```
liljonnystyle, line 96: this should probably call the `PowerDemandModel()`

flow/core/kernel/vehicle/traci.py
```diff
@@ -5,15 +5,20 @@
...
 import numpy as np
+import math
 import collections
 import warnings
+import random
+import statistics
 from flow.controllers.car_following_models import SimCarFollowingController
 from flow.controllers.rlcontroller import RLController
 from flow.controllers.lane_change_controllers import SimLaneChangeController
+from flow.core.kernel.energy import Energy
```
liljonnystyle, line 16: import the correct names (all of the classes except the base class)

flow/core/kernel/vehicle/traci.py
```diff
@@ -263,6 +281,8 @@ def update(self, reset):
         # make sure the rl vehicle list is still sorted
         self.__rl_ids.sort()
 
+        self.energy.calc_energy_level(self)
```
liljonnystyle, line 284: this method doesn't exist

flow/core/kernel/vehicle/traci.py
```diff
@@ -536,13 +556,6 @@ def get_num_not_departed(self):
         """See parent class."""
         return self.num_not_departed
 
-    def get_fuel_consumption(self, veh_id, error=-1001):
-        """Return fuel consumption in gallons/s."""
-        ml_to_gallons = 0.000264172
-        if isinstance(veh_id, (list, np.ndarray)):
-            return [self.get_fuel_consumption(vehID, error) for vehID in veh_id]
-        return self.__sumo_obs.get(veh_id, {}).get(tc.VAR_FUELCONSUMPTION, error) * ml_to_gallons
-
```
liljonnystyle, line 545: if we strip this out, we need to know what to plug in in its place. @Yasharzf and/or Sulaiman can help

flow/core/kernel/vehicle/traci.py
```diff
@@ -1120,3 +1133,17 @@ def get_max_speed(self, veh_id, error=-1001):
...
+    def set_soc(self, veh_id, soc):
+        """Set the lane headways of the specified vehicle."""
+        self.__vehicles[veh_id]['SOC'] = soc
+
+    def get_fuel(self, veh_id):
+        return self.__vehicles[veh_id]['fuel']
+
+    def set_fuel(self, veh_id, fuel):
+        """Set the lane headways of the specified vehicle."""
+        self.__vehicles[veh_id]['fuel'] = fuel
```
liljonnystyle, line 1149: not sure if this is the best place for these, since soc and fuel are mutually exclusive. it would be nice if (at least) they could be consolidated to something using consistent units.

flow/core/kernel/energy.py
```diff
@@ -0,0 +1,119 @@
...
+class EnergyModel():
+    """Base energy model class."""
+
+    def __init__(self,kernel):
+        self.vehicle = vehicle
+
+    def get_energy(self, veh_id):
+        #return self.calc_energy(veh_id)
+        pass
+
```
liljonnystyle, line 21: i think this might be a good place to implement common unit conversions that will be used by the child classes, like joules <-> megajoules, soc <-> joules, gallons <-> joules, etc.

flow/core/kernel/base_energy.py
```diff
@@ -0,0 +1,32 @@
...
+    """Base energy model class.
+
+    Calculate the instantaneous power consumption of a vehicle in
+    the network.  It returns the power in Watts regardless of the 
+    vehicle type: whether EV or Combustion Engine, Toyota Prius or Tacoma
+    or non-Toyota vehicles. Non-Toyota vehicles are set by default 
+    to be an averaged-size vehicle.
+    """
+
+    def __init__(self,kernel):
```
liljonnystyle, line 23: add space after all commas

flow/core/kernel/base_energy.py
```diff
@@ -0,0 +1,32 @@
...
+        self.vehicle = vehicle
+
+    @abstractmethod
+    def get_instantaneous_power(self, veh_id, model_param, grade):
```
liljonnystyle, line 27: i've renamed this function, so change accordingly below.

since you've instantiated `self.vehicle` i don't think you need `veh_id` anymore. in fact, you should probably also instantiate the `env` in the `__init__` so that you don't need to supply the grade or other environment variables here (like speed, acceleration).

flow/core/kernel/power_demand.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+    def __init__(self,kernel):
+        self.k = kernel
+        self.g = 9.8
+        self.rho_air = 1.225
+        self.Mass = 1200
+        self.rolling_res_coeff = 0.005
+        self.aerodymnamic_drag_coeff = 0.3
+        self.cross_A = 2.6
```
liljonnystyle, line 23: - make g = 9.81 to be consistent with some other code we have
- all variable names should be lower case (i.e. `self.mass` and `self.cross_area`)
- typo in 'aerodynamic'
- let's include `self.gamma = 1` here

flow/core/kernel/power_demand.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def get_energy(self, veh_id, grade):
+        speed = self.k.get_speed(veh_id)
+        if veh_id in self.k.previous_speeds:
+            old_speed = self.k.previous_speeds[veh_id]
+        else:
+            old_speed = speed
+
+        accel = (speed - old_speed)/self.sim_step
+        power = self.Mass*speed*(accel+self.g*math.sin(grade))+self.Mass*self.g*self.rolling_res_coeff*speed+0.5*self.rho_air*self.cross_A*self.aerodymnamic_drag_coeff*speed**3
```
liljonnystyle, line 41: rewrite this in multiple lines and spaces in between operators, so that it's easier to read. also include the heaviside function here.

i might even consider writing this as three separate terms first, i.e.:
```
accel_slope_forces = self.mass * speed * (
                     (np.heaviside(accel, 0.5) * (1 - self.gamma) + self.gamma)) * accel
                     + self.g * math.sin(grade))
rolling_friction = self.mass * self.g * self.rolling_res_coeff * speed
air_drag = 0.5 * self.rho_air * self.cross_area * self.aerodynamic_drag_coeff * speed**3

power = accel_slope_forces + rolling_friction + air_drag
```
but please check that i've written this correctly!

flow/core/kernel/power_demand.py
```diff
@@ -0,0 +1,60 @@
...
+        power = max(0,power)
```
liljonnystyle, line 58: since everything above this line is identical in the PDM_CEV, let's pull everything out into a separate function in the parent class (like `general_power_demand(self, veh_id, grade)`) and then call that function within this function.

flow/core/kernel/power_demand.py
```diff
@@ -0,0 +1,60 @@
...
+        self.Mass = 1200
+        self.rolling_res_coeff = 0.005
+        self.aerodymnamic_drag_coeff = 0.3
+        self.cross_A = 2.6
+    
+    def get_energy(self, veh_id, grade):
+        pass
+    
+
+class PDM_CEV(PowerDemandModel):
```
liljonnystyle, line 29: class naming convention is to use CamelCase without underscores, so let's call this `PDMCombustionEngine` and the one below `PDMElectric`

flow/core/kernel/toyota_energy.py
```diff
@@ -0,0 +1,71 @@
...
+        # delete pickle file
+        os.remove(file.pkl)
+
+    def get_energy(self,veh_id,soc,grade):
+        pass
+
+
+class prius_energy(ToyotaModel):
+
+    super.__init__(kernel,filename = 'prius_test.pkl')
```
liljonnystyle, line 32: sorry, i had this slightly wrong. this should be:
```
def __init__(self, kernel):
    super(PriusEnergy, self).__init__(kernel, filename = 'prius_test.pkl')
```
and I think all the parent/grandparent classes above should have the same call signature (i.e. `def __init__(self, kernel, filename=None):`

flow/core/kernel/toyota_energy.py
```diff
@@ -0,0 +1,71 @@
...
+        with open('file.pkl','rb')as file:
+            self.toyota_energy = pickle.load(file) #self.prius_energy
+        # delete pickle file
+        os.remove(file.pkl)
+
+    def get_energy(self,veh_id,soc,grade):
+        pass
+
+
+class prius_energy(ToyotaModel):
```
liljonnystyle, line 30: `class PriusEnergy(ToyotaModel):`

flow/core/kernel/toyota_energy.py
```diff
@@ -0,0 +1,71 @@
...
+        accel = (speed - old_speed)/self.sim_step
+
+        old_soc = self.k.get_soc(veh_id)
+        grade = 0 
+
+        socdot = self.toyota_energy(old_soc,speed,accel,grade)
+
+        return socdot
+            
+class tacoma_energy(ToyotaModel):
```
liljonnystyle, line 52: `class TacomaEnergy(ToyotaModel):`

flow/core/kernel/toyota_energy.py
```diff
@@ -0,0 +1,71 @@
...
+
+        if veh_id in self.k.previous_speeds:
+            old_speed = self.k.previous_speeds[veh_id]
+        else:
+            old_speed = speed
+
+        accel = (speed - old_speed)/self.sim_step
+        grade = 0 
+
+        fc = tacoma_test(speed,accel,grade) # returns instantanoes fuel consumption
```
liljonnystyle, line 68: `self.toyota_energy(`...

flow/core/kernel/vehicle/traci.py
```diff
@@ -5,15 +5,20 @@
...
 import numpy as np
+import math
 import collections
 import warnings
+import random
+import statistics
 from flow.controllers.car_following_models import SimCarFollowingController
 from flow.controllers.rlcontroller import RLController
 from flow.controllers.lane_change_controllers import SimLaneChangeController
+from flow.core.kernel.base_energy import BaseEnergyModel
```
liljonnystyle, line 16: import all the energy classes here

flow/core/kernel/vehicle/traci.py
```diff
@@ -88,6 +93,12 @@ def __init__(self,
         # old speeds used to compute accelerations
         self.previous_speeds = {}
 
+        self.energy = Energy(self)
+        self.energy.calc_energy_level(self)
+
+        # # store speeds of all vehicles at last iteration
+        self.old_speeds = {}
+
```
liljonnystyle, line 101: I think this was leftover from some old code, you can probably delete these lines.

flow/core/kernel/vehicle/traci.py
```diff
@@ -263,6 +283,8 @@ def update(self, reset):
         # make sure the rl vehicle list is still sorted
         self.__rl_ids.sort()
 
+        #self.base_energy.get_energy(self)
+
```
liljonnystyle, line 287: delete this if you don't need it

flow/core/kernel/base_energy.py
```diff
@@ -0,0 +1,23 @@
...
+
+    Calculate the instantaneous power consumption of a vehicle in
+    the network.  It returns the power in Watts regardless of the 
+    vehicle type: whether EV or Combustion Engine, Toyota Prius or Tacoma
+    or non-Toyota vehicles. Non-Toyota vehicles are set by default 
+    to be an averaged-size vehicle.
+    """
+
+    def __init__(self, kernel):
+        self.k.env.vehicle = vehicle
```
liljonnystyle, line 15: we'll need to figure out what goes here specifically based on how these classes will be instantiated. follow up on Eugene's suggestions, and let's talk to Sulaiman/Yashar on this.

i think `filename` should be an input here since the Toyota ones need it, so you can just add `filename=None` here.

also, for the vehicle, you are currently assigning `vehicle` to `self.k.env.vehicle`, but there are a couple problems with this.
1. vehicle is an unknown object, unless you add it to the call signature
2. you can't assign anything to `self.k.env.vehicle` until it knows what `self.k.env` is. similarly, you can't assign anything to `self.k.env` until it knows what `self.k` is.

so i think what you'll want to do is something like:
```
def __init__(self, kernel, vehicle, filename=None):
    self.kernel = kernel # i suspect you don't need this, but i'm not sure yet
    self.vehicle = vehicle
```

flow/core/kernel/power_demand.py
```diff
@@ -0,0 +1,64 @@
...
+class PowerDemandModel(BaseEnergyModel, metaclass=ABCMeta):
+    """Calculate power consumption of a vehicle.
+    Assumes vehicle is an average sized vehicle.
+    The power calculated here is the lower bound of the 
+    actual power consumed by the vehicle
+    """
+
+    def __init__(self,kernel):
+        self.k = kernel
+        self.k.env.vehicle = vehicle
```
liljonnystyle, line 19: At the top of these child init's, you can inherit the parent class init by doing:
`super(BaseEnergyModel, self).__init__(<fill in call signature here>)`

flow/core/kernel/power_demand.py
```diff
@@ -0,0 +1,64 @@
...
+    def get_instantaneous_power(self, veh_id, grade):
+        pass
+    
+
+class PDMCombustionEngine(PowerDemandModel):
+    
+    # Power Demand Model for a combustion engine vehicle
+
+    def get_instantaneous_power(self, veh_id, grade):
+        power = self.calculate_power
```
liljonnystyle, line 54: this is a function: `self.calculate_power()`

flow/core/kernel/power_demand.py
```diff
@@ -0,0 +1,64 @@
...
+        return power
+
+class PDMElectric(PowerDemandModel):
+    
+    # Power Demand Model for an electric vehicle
+
+    def get_instantaneous_power(self, veh_id, grade):
+        power = max(0,self.calculate_power)
```
liljonnystyle, line 62: same here

flow/core/kernel/toyota_energy.py
```diff
@@ -0,0 +1,76 @@
...
+from collections import namedtuple
+from flow.energy.base_energy import BaseEnergyModel
+import os
+from abc import ABCMeta, abstractmethod
+
+class ToyotaModel(BaseEnergyModel):
+
+    def __init__(self, kernel, filename=None):
+        self.k = kernel
+        self.k.env.vehicle = vehicle
```
liljonnystyle, line 18: `super(BaseEnergyModel, ...`

flow/core/kernel/vehicle/traci.py
liljonnystyle: `PowerDemandModel, PDMCombustion,` etc.

flow/core/kernel/vehicle/traci.py
liljonnystyle: i think the instantaneous power is not a function of fuel, right? so we don't need to track it (do we?).

for SOC, maybe we should just pick an estimated value, like 0.5, but track it inside the `ToyotaPriusModel`, not out here.

flow/core/kernel/vehicle/traci.py
```diff
@@ -1195,3 +1217,17 @@ def get_road_grade(self, veh_id):
...
+    def set_soc(self, veh_id, soc):
+        """Set the lane headways of the specified vehicle."""
+        self.__vehicles[veh_id]['SOC'] = soc
+
+    def get_fuel(self, veh_id):
+        return self.__vehicles[veh_id]['fuel']
+
+    def set_fuel(self, veh_id, fuel):
+        """Set the lane headways of the specified vehicle."""
+        self.__vehicles[veh_id]['fuel'] = fuel
```
liljonnystyle, line 1233: see above comment, i think we can delete these.

flow/core/kernel/toyota_energy.py
```diff
@@ -0,0 +1,76 @@
...
+            old_speed = speed
+
+        accel = (speed - old_speed)/self.k.env.sim_step
+
+        old_soc = self.k.env.get_soc(veh_id)
+        grade = 0 
+
+        socdot = self.toyota_energy(old_soc, speed, accel, grade)
+
+        return socdot
```
liljonnystyle, line 54: we still need to convert this to Watts

flow/core/kernel/toyota_energy.py
```diff
@@ -0,0 +1,76 @@
...
+        if veh_id in self.k.env.previous_speeds:
+            old_speed = self.k.env.previous_speeds[veh_id]
+        else:
+            old_speed = speed
+
+        accel = (speed - old_speed)/self.k.env.sim_step
+        grade = 0 
+
+        fc = self.toyota_energy(speed,accel,grade) # returns instantanoes fuel consumption
+        return fc
```
liljonnystyle, line 74: does this return Watts or mL or what? make sure we are outputting Watts

flow/core/kernel/base_energy.py
Yasharzf: you want to call methods from other sub-kernels (vehicle). to be consistent, you can use "master_kernel" similar to what we have in the other sub-kernel base classes.

flow/core/kernel/power_demand.py
```diff
@@ -16,49 +16,38 @@ class PowerDemandModel(BaseEnergyModel, metaclass=ABCMeta):
...
-            old_speed = self.k.env.previous_speeds[veh_id]
-        else:
-            old_speed = speed
-
-        accel = (speed - old_speed)/self.k.env.sim_step
-        accel_slope_forces = self.mass * speed * (
-                     (np.heaviside(accel, 0.5) * (1 - self.gamma) + self.gamma)) * accel
-                     + self.g * math.sin(grade))
+    def calculate_power(self, accel, speed, grade):
+        accel_slope_forces = self.mass * speed * ((np.heaviside(accel, 0.5) * (1 - self.gamma) + self.gamma)) * accel + self.g * math.sin(grade)
```
liljonnystyle, line 28: i think the parens should be like this:
`self.gamma) * accel + self.g * math.sin(grade))`

also this line is too long, so reformat it into 2-3 lines

flow/core/kernel/base_energy.py
```diff
@@ -12,10 +11,10 @@ class BaseEnergyModel(metaclass=ABCMeta):
     """
 
     def __init__(self, kernel):
-        self.k.env.vehicle = vehicle
+        self.k = kernel
 
     @abstractmethod
-    def get_instantaneous_power(self):
+    def get_instantaneous_power(self, parameter, accel, speed, grade):
```
liljonnystyle, line 17: it looks like `parameter` is for the old_soc, and i guess you are saving this for any other future parameters you might want to pass? if so, i think the pythonic way to do this is using args and kwargs (i'm not super familiar but you can read about it here: https://book.pythontips.com/en/latest/args_and_kwargs.html)

you should put this parameter last, and have it default to None if it's not passed.

flow/core/rewards.py
```diff
@@ -441,3 +441,52 @@ def miles_per_gallon(env, veh_ids=None, gain=.001):
...
+
+    if veh_ids is None:
+        veh_ids = env.k.vehicle.get_ids()
+    elif not isinstance(veh_ids, list):
+        veh_ids = [veh_ids]
+    for veh_id in veh_ids:
+        speed = env.k.vehicle.get_speed(veh_id)
+        prev_speed = env.k.vehicle.get_previous_speed(veh_id)
+        # negative acceleration indicates regenerative breaking for some energy models
+        accel = (speed - prev_speed) / env.sim_step
```
liljonnystyle, line 474: for accel, you can use:
`env.k.vehicle.get_accel_no_noise_with_failsafe(veh_id)`
then you don't need prev_speed anymore

flow/core/rewards.py
```diff
@@ -441,3 +441,52 @@ def miles_per_gallon(env, veh_ids=None, gain=.001):
...
+        if veh_id is "Prius":
+            energy_model = env.k.PriusEnergy
+            initial_soc = 0.9
+            parameter = initial_soc
+        elif veh_id is "Tacoma":
+            energy_model = env.k.TacomaEnergy
+        elif veh_id is "EV":
+            energy_model = env.k.PDMElectric
+        else:
+            energy_model = env.k.PDMCombustionEngine
```
liljonnystyle, line 488: maybe don't worry about the dictionary lookup here. but for each condition, use this syntax:
`if 'prius' in veh_id:`
because the veh_id's will be like "av_prius" or "human_prius".

also, instead of `else` on the last condition, use `elif` so that we can still catch anything that doesn't fall properly into one of the 4 (unless you want `PDMCombustionEngine` as the default, then ok).

flow/core/rewards.py
```diff
@@ -441,3 +441,52 @@ def miles_per_gallon(env, veh_ids=None, gain=.001):
     mpg /= 1609.0
 
     return mpg * gain
+
+
+def instantaneous_power(env, veh_ids=None, gain=.001):
```
liljonnystyle, line 446: i don’t think we want to re-instantiate the class every time this function is called, so you should find a way to instantiate it at the beginning of the simulation, at the level of VehicleParams

flow/core/rewards.py
```diff
@@ -441,3 +441,52 @@ def miles_per_gallon(env, veh_ids=None, gain=.001):
...
+        prev_speed = env.k.vehicle.get_previous_speed(veh_id)
+        # negative acceleration indicates regenerative breaking for some energy models
+        accel = (speed - prev_speed) / env.sim_step
+        grade = 0
+
+        parameter = None
+        # determine which energy model to use
+        if veh_id is "Prius":
+            energy_model = env.k.PriusEnergy
+            initial_soc = 0.9
```
liljonnystyle, line 481: initial_soc is always reset to 0.9 every time this function is called

joyncarpio: Should we let the user set the initial soc instead?

flow/core/kernel/toyota_energy.py
```diff
@@ -0,0 +1,51 @@
...
+    @abstractmethod
+    def get_instantaneous_power(self):
+        pass
+
+
+class PriusEnergy(ToyotaModel):
+    
+    def __init__(self, kernel):
+        super(PriusEnergy, self).__init__(kernel, filename = 'prius_test.pkl')
+
```
liljonnystyle, line 36: self.soc = initial_soc

flow/core/kernel/toyota_energy.py
```diff
@@ -0,0 +1,51 @@
...
+    def get_instantaneous_power(self, parameter, accel, speed, grade):
+
+        socdot = self.toyota_energy(parameter, accel, speed, grade)
```
liljonnystyle, line 39: socdot = self.toyota_energy(self.soc, accel, speed, grade)

then update self.soc

flow/core/params.py
```diff
@@ -8,6 +8,14 @@
...
 from flow.controllers.lane_change_controllers import SimLaneChangeController
 
+# To add these after creating the forlder for energy models # (1)
+"""
+from flow.enermod.toyota_energy import PriusEnergy
+from flow.enermod.toyota_energy import TacomaEnergy
+from flow.enermod.power_demand import PDMCombustionEngine
+from flow.enermod.power_demand import PDMElectric
+"""
+
```
joyncarpio, line 18: Plan is to create a folder consisting of all the energy models. Temporarily i name it as "enermod".  This section is commented for now until the idea is approved and the folder has been created.

flow/core/params.py
```diff
@@ -270,6 +278,19 @@ def add(self,
...
+            energy_model = env.k.PriusEnergy.get_instantaneous_power()
+        elif "tacoma" in veh_id:
+            energy_model = env.k.TacomaEnergy.get_instantaneous_power()
+        elif "ev" in veh_id:
+            energy_model = env.k.PDMElectric.get_instantaneous_power()
+        else:
+            energy_model = env.k.PDMCombustionEngine.get_instantaneous_power()
+   
+    
+
```
joyncarpio, line 293: This is where assignment of energy models happen inside the VehicleParams().  I set PDMCombustionEngine model as the default energy model.

flow/core/rewards.py
```diff
@@ -442,8 +442,7 @@ def miles_per_gallon(env, veh_ids=None, gain=.001):
 
     return mpg * gain
 
-
-def instantaneous_power(env, veh_ids=None, gain=.001):
+def instantaneous_power(env, veh_ids=None, parameter=None, gain=.001):
```
joyncarpio, line 445: I added parameter as an argument of this function in the rewards.py, I will apply the *args / **kwargs in replacement for this parameter.  But I am uncertain, how to pass this parameter value to the initialization of the energy models since I will only be needing this value when I am going to set initial_soc.

flow/core/kernel/toyota_energy.py
```diff
@@ -25,18 +25,19 @@ def __init__(self, kernel, filename=None):
...
-    def get_instantaneous_power(self):
+    def get_instantaneous_power(self, accel, speed, grade):
         pass
 
 
 class PriusEnergy(ToyotaModel):
     
     def __init__(self, kernel):
         super(PriusEnergy, self).__init__(kernel, filename = 'prius_test.pkl')
+        self.soc = # how to pass here the parameter vargument from the rewards.py
```
joyncarpio, line 36: This is where i should use the value passed in the "parameter" in the function in rewards.py.

liljonnystyle: so i think this is where you use the `args` or `kwargs` in the `__init__` call signature to get `initial_soc`.

you'll just initialize it in VehicleParams (with a default value of 0.9, for example). you won't need to pass a new `soc` every time from rewards.py because this class should be tracking/updating the soc at every time_step

flow/core/kernel/toyota_energy.py
liljonnystyle: remember to pass `self.soc` instead of `parameter`

and in the line below, i believe you need `self.k.env.sim_step` or something? and typically, we do updates like this:
`self.soc += socdot * self.k.env.sim_step`
(also, check on the sign of socdot)

flow/core/params.py
liljonnystyle: yes this is fine. let's name the folder `energy_models`

flow/core/params.py
liljonnystyle: rather than instantiating the instantaneous power method here, you should just provide the class names, e.g.
```
if "prius" in veh_id:
    energy_class_name = PriusEnergy
```

when the actual vehicles are added to the simulation domain (i think in `add_departed()`?), there you will instantiate the class. i'm not sure about the design choice, but you might have a dictionary of energy classes (say `energy_classes = dict()`), and every time a new vehicle is added, you say something like:
```
energy_classes[veh_id] = energy_class_name(kernel, **kwargs)
```
and this is where you need to figure out the args/kwargs stuff, to pass `initial_soc` if needed.

flow/core/rewards.py
liljonnystyle: see above note, so i believe you won't need parameter or gain here anymore.

flow/core/rewards.py
```diff
@@ -469,24 +468,10 @@ def instantaneous_power(env, veh_ids=None, gain=.001):
         veh_ids = [veh_ids]
     for veh_id in veh_ids:
         speed = env.k.vehicle.get_speed(veh_id)
-        prev_speed = env.k.vehicle.get_previous_speed(veh_id)
-        # negative acceleration indicates regenerative breaking for some energy models
-        accel = (speed - prev_speed) / env.sim_step
+        accel = env.k.vehicle.get_accel_no_noise_with_failsafe(veh_id)
         grade = 0
```
liljonnystyle, line 472: you can also use `env.k.vehicle.get_road_grade(veh_id)` here, which currently returns 0. this way, this will automatically work when that method is filled in.

flow/core/rewards.py
```diff
@@ -469,24 +468,10 @@ def instantaneous_power(env, veh_ids=None, gain=.001):
...
-        elif veh_id is "Tacoma":
-            energy_model = env.k.TacomaEnergy
-        elif veh_id is "EV":
-            energy_model = env.k.PDMElectric
-        else:
-            energy_model = env.k.PDMCombustionEngine
-        
-        inst_power = energy_model.get_instantaneous_power(parameter, accel, speed, grade)
+        # need to pass the parameter value to the init of the energy models ?????
+        inst_power = energy_model(accel, speed, grade)
```
liljonnystyle, line 474: so when you get here, the class should already be instantiated for the vehicle, and you just need to access the method within the class, so:
`inst_power = energy_classes[veh_id].get_instantaneous_power(accel, speed, grade)`

flow/core/kernel/vehicle/traci.py
```diff
@@ -5,15 +5,23 @@
...
 import warnings
+import random
+import statistics
 from flow.controllers.car_following_models import SimCarFollowingController
 from flow.controllers.rlcontroller import RLController
 from flow.controllers.lane_change_controllers import SimLaneChangeController
+from flow.energy_models.toyota_energy import PriusEnergy
+from flow.energy_models.toyota_energy import TacomaEnergy
+from flow.energy_models.power_demand import PDMCombustionEngine
+from flow.energy_models.power_demand import PDMElectric
```
liljonnystyle, line 19: I think all these new imports can be removed since they are not used here (math, random, statistics, energy classes)

flow/core/kernel/vehicle/traci.py
```diff
@@ -113,6 +127,15 @@ def initialize(self, vehicles):
...
                 self.__vehicles[veh_id]['initial_speed'] = typ['initial_speed']
+
+                # self.energy.initialize(veh_id)
+                # energy.intialize(veh_id)
+                # self.__vehicles[veh_id]['SOC'] = float(random.randrange(8, 92))/100 #move to experiment code
+                # # Assume level of fuel is 10-100% of capacity
+                # # Tacoma has 21.1 gallon capacity = 79.87219 L
+                # # Use density of gasoline = 748.9 g/L (needs to convert b/c fc is in g/s)
+                # self.__vehicles[veh_id]['fuel'] = float(random.randrange(598163, 5981628))/100
+
```
liljonnystyle, line 138: remove

flow/core/kernel/vehicle/traci.py
```diff
@@ -1195,3 +1223,19 @@ def get_road_grade(self, veh_id):
...
+        self.__vehicles[veh_id]['SOC'] = soc
+
+    def get_fuel(self, veh_id):
+        return self.__vehicles[veh_id]['fuel']
+
+    def set_fuel(self, veh_id, fuel):
+        """Set the lane headways of the specified vehicle."""
+        self.__vehicles[veh_id]['fuel'] = fuel
+    """
+
```
liljonnystyle, line 1241: remove

flow/core/params.py
```diff
@@ -270,6 +276,27 @@ def add(self,
...
+        if energy_model is "prius":
+            self.energy_model = PriusEnergy(kernel, 0.9)
+        elif energy_model is "tacoma":
+            self.energy_model = TacomaEnergy()
+        elif energy_model is "ev":
+            self.energy_model = PDMElectric()
+        else:
+            self.energy_model = PDMCombustionEngine()
+
```
liljonnystyle, line 299: i don't think we need this anymore. follow the examples below for acceleration_controller (see `self.type_parameters`). also for reference, notice how controller classes are provided in `vehicles.add()`:
```
vehicles.add(
    "human",
    num_vehicles=0,
    lane_change_params=SumoLaneChangeParams(
        lane_change_mode="strategic",
    ),
    acceleration_controller=(IDMController, {
        "a": 1.3,
        "b": 2.0,
        "noise": 0.3,
    }),
    routing_controller=(I210Router, {}) if ON_RAMP else None,
)
```

flow/core/rewards.py
```diff
@@ -441,3 +441,35 @@ def miles_per_gallon(env, veh_ids=None, gain=.001):
...
+    veh_ids : [list]
+        list of veh_ids to compute the reward over
+    gain : float
+        scaling factor for the reward
+    """
+
+    if veh_ids is None:
+        veh_ids = env.k.vehicle.get_ids()
+    elif not isinstance(veh_ids, list):
+        veh_ids = [veh_ids]
```
liljonnystyle, line 468: add a blank line between `if` logic and `for` loop for readability

flow/energy_models/power_demand.py
```diff
@@ -0,0 +1,54 @@
...
+        air_drag = 0.5 * self.rho_air * self.cross_area * self.aerodynamic_drag_coeff * speed**3
+        power = accel_slope_forces + rolling_friction + air_drag
+        return power
+      
+    @abstractmethod
+    def get_regen_cap(self, accel, speed, grade)
+        pass
+
+    def get_instantaneous_power(self, accel, speed, grade):
+        power = max(get_regen_cap(), self.calculate_power(accel, speed, grade))
```
liljonnystyle, line 40: `self.get_regen_cap()`

flow/energy_models/power_demand.py
```diff
@@ -0,0 +1,54 @@
...
+class PDMCombustionEngine(PowerDemandModel):
+    
+    # Power Demand Model for a combustion engine vehicle
+    def get_regen_cap(self, accel, speed, grade)
+        return 0
+
+class PDMElectric(PowerDemandModel):
+    
+    # Power Demand Model for an electric vehicle
+    def get_regen_cap(self, accel, speed, grade)
```
liljonnystyle, line 53: missing `:` at the end of these function def lines

flow/core/rewards.py
```diff
@@ -441,3 +441,35 @@ def miles_per_gallon(env, veh_ids=None, gain=.001):
...
+    for veh_id in veh_ids:
+        speed = env.k.vehicle.get_speed(veh_id)
+        accel = env.k.vehicle.get_accel_no_noise_with_failsafe(veh_id)
+        grade = env.k.vehicle.get_road_grade(veh_id)
+        inst_power = energy_model[veh_id].get_instantaneous_power(accel, speed, grade)
```
liljonnystyle, line 473: this call structure will probably change depending on how you change the stuff above

flow/core/params.py
```diff
@@ -242,7 +247,8 @@ def add(self,
             num_vehicles=0,
             car_following_params=None,
             lane_change_params=None,
-            color=None):
+            color=None,
+            energy_model = None):
```
liljonnystyle, line 251: minor: i would put `energy_model` above `color` since it is kind of the same "level" as the upper args.

flow/energy_models/power_demand.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __init__(self,kernel):
+        self.k = kernel
+        self.g = 9.81
+        self.rho_air = 1.225
+        self.mass = 1200
+        self.rolling_res_coeff = 0.005
+        self.aerodynamic_drag_coeff = 0.3
+        self.cross_area = 2.6
+        self.gamma = 1
```
liljonnystyle, line 25: some of these numbers are specific to prius (move to PDMElectric), and some need to be modified for PDMCombustionEngine

flow/core/kernel/vehicle/traci.py
```diff
@@ -554,6 +585,12 @@ def get_speed(self, veh_id, error=-1001):
         if isinstance(veh_id, (list, np.ndarray)):
             return [self.get_speed(vehID, error) for vehID in veh_id]
         return self.__sumo_obs.get(veh_id, {}).get(tc.VAR_SPEED, error)
+    
+    def get_power(self, veh_id, error=-1001):
+        """See parent class."""
+        if isinstance(veh_id, (list, np.ndarray)):
+            return [self.get_power(vehID, error) for vehID in veh_id]
+        return self.__vehicles.get(veh_id, {}).get("energy_model", error)
```
liljonnystyle, line 593: I think we need to create this as an abstractmethod in the base vehicle class with a proper docstring. Let's rename this to `get_energy_model` to be a bit more descriptive.

Also, I think if this returns `error`, then later it will break if we try to call `error.get_instantaneous_power()`

flow/energy_models/power_demand.py
```diff
@@ -0,0 +1,57 @@
...
+from abc import ABCMeta, abstractmethod
+
+class PowerDemandModel(BaseEnergyModel, metaclass=ABCMeta):
+    """Calculate power consumption of a vehicle.
+    Assumes vehicle is an average sized vehicle.
+    The power calculated here is the lower bound of the 
+    actual power consumed by the vehicle
+    """
+
+    def __init__(self, kernel, mass=2041, area=3.2, Cr=0.0027, Ca=0.4):
```
liljonnystyle, line 17: We probably want these parameters to be lower-case variables for consistent styling purposes. you may as well just call them `rolling_res_coeff` and `aerodynamic_drag_coeff`

flow/core/rewards.py
```diff
@@ -441,3 +441,35 @@ def miles_per_gallon(env, veh_ids=None, gain=.001):
...
     return mpg * gain
+
+def instantaneous_power(env, veh_ids=None, gain=.001):
+    """Calculate the instantaneous power for every simulation step specific to the vehicle type.
+
+    The energy model used is based on the vehicle type:
+    veh_id = "Prius"  --> energy model is PriusEnergy in flow.core.kernel.toyata_energy
+    veh_id = "Tacoma" --> energy model is TacomaEnergy in flow.core.kernel.toyata_energy
+    veh_id = "EV"     --> energy model is PDMElectric in flow.core.kernel.power_demand
+    veh_id = "human"  --> energy model is PDMCombustionEngine in flow.core.kernel.power_demand
```
liljonnystyle, line 452: update the docstring accordingly

flow/core/params.py
```diff
@@ -242,7 +247,8 @@ def add(self,
             num_vehicles=0,
             car_following_params=None,
             lane_change_params=None,
-            color=None):
+            color=None,
+            energy_model = (PDMCombustionEngine, {})):
```
liljonnystyle, line 251: nit: put `energy_model` above `color`

flow/core/kernel/vehicle/traci.py
kjang96: +1 to all of the above

flow/core/params.py
kjang96: Remove the spaces around the =

flow/core/rewards.py
kjang96: Change `veh_ids` in the function header to `veh_id`. This is so we can match the getter functions in the flow/core/kernel/vehicle/traci.py . If you could make the changes to the `miles_per_gallon` function in rewards.py, that would be awesome

flow/energy_models/base_energy.py
```diff
@@ -0,0 +1,22 @@
+from abc import ABCMeta, abstractmethod
+
+class BaseEnergyModel(metaclass=ABCMeta):
```
kjang96, line 3: I don't think you need to have this ABCMeta thing. I'm a little rusty on metaclasses and might be wrong though, so first, my question for you is -- why did you put this here?

liljonnystyle: ah this was my suggestion. i made a bunch of the base classes throughout the codebase into abstract base classes (that was landed in master, not sure if it trickled down to i210_dev). basically i wanted any instantiation of a base class to error right away, rather than erroring when an un-implemented method within is called.

flow/core/rewards.py
```diff
@@ -441,3 +441,29 @@ def miles_per_gallon(env, veh_ids=None, gain=.001):
     mpg /= 1609.0
 
     return mpg * gain
+
+def instantaneous_power(env, veh_id=None, gain=.001):
```
liljonnystyle, line 445: with this changed to `veh_id` then the docstring and function are no longer valid. i think the plan was to just simplify this function for handling one vehicle at a time? i'm not 100% sure.

flow/core/params.py
```diff
@@ -304,6 +306,9 @@ def add(self,
             # FIXME: depends on simulator
             lane_change_params = SumoLaneChangeParams()
 
+        if energy_model not in ENERGY_MODELS:
```
kjang96, line 309: I think here we should also add a print statement like print([energy model] is not currently supported)

flow/energy_models/power_demand.py
```diff
@@ -1,57 +1,127 @@
+"""Script containing the vehicle power demand model energy classes."""
```
kjang96, line 1: Just curious where did all these models come from?

flow/energy_models/toyota_energy.py
```diff
@@ -1,50 +1,58 @@
...
-        super(PriusEnergy, self).__init__(kernel, filename = 'prius_test.pkl')
-        self.soc = soc 
+        super(PriusEnergy, self).__init__(kernel, filename='prius_ev.pkl')
+        self.soc = soc
 
     def get_instantaneous_power(self, accel, speed, grade):
+        """See parent class."""
         socdot = self.toyota_energy(self.soc, accel, speed, grade)
         self.soc -= socdot * self.k.env.sim_step
+        # FIXME (Joy): convert socdot to power
```
kjang96, line 41: Are we taking care of this while Joy's on vacation?

flow/energy_models/power_demand.py
liljonnystyle: The physics-based model that we got from Benni:
image

flow/energy_models/toyota_energy.py
liljonnystyle: My plan is to leave it unfixed since it isn't really part of the MVP

flow/core/kernel/vehicle/traci.py
17150934: How to add an electric vehicle and get ElectricityConsumption

PR #948, opened 2020-05-27
Update lane change mode

Yasharzf:
Existing lane change modes are ambiguous and misleading. For instance, "strategic" lane change mode description says "Human cars make lane changes in accordance with SUMO to provide speed boosts" however, the assigned bitset is 1621 which is SUMO's default lane change mode (execute all changes unless in conflict with TraCI). Or the bitset for "aggressive" mode is 0 which in addition to deactivating safety checks, deactivates all lane changes. I added additional lane change modes.


flow/core/params.py
```diff
@@ -897,14 +917,71 @@ class SumoLaneChangeParams:
     ----------
     lane_change_mode : str or int, optional
         may be one of the following:
+        * "no_lc_safe" (default): Disable all autonomous changing but still
+          handle safety checks (collision avoidance and safety-gap enforcement)
+          in the simulation. Binary is [001000000000]
```
eugenevinitsky, line 922: We need to be careful about the word autonomous, this implies to me that the AVs cannot lane change. I think it's worth expanding the text here to be clearer about the effects.

flow/core/params.py
```diff
@@ -897,14 +917,71 @@ class SumoLaneChangeParams:
...
+        * "no_lc_aggressive": SUMO lane changes are not executed, collision
+          avoidance and safety-gap enforcement are off.
+          Binary is [000000000000]
+
+        * "sumo_default": Execute all changes unless in conflict with TraCI.
+          Binary is [011001010101].
```
eugenevinitsky, line 928: nit: can you make clear what "all" refers to?

flow/core/params.py
eugenevinitsky: I think if you say "disable all SUMO lane changing" that might be good

PR #955, opened 2020-05-29
Done fix

eugenevinitsky:
The done condition is not correct for sims_per_step greater than one. Its possible a vehicle might have left during one of the steps. Additionally, fix the done condition to get set when a vehicle is on the exit edge.


eugenevinitsky:
I'm not sure why this isn't getting the histories right, this is built on top of I210 dev...

PR #967, opened 2020-06-12
add emergencydecel_warning_threshold parameter

pengyuan-zhou:
Hi, 

I found it bothered the tons of emergency braking warnings when running large scale tests.
So added this parameter that one can set its value as 1.1 in example file to disable these warnings.

Cheers,
Pengyuan


pengyuan-zhou:
The parameter causes an issue, will be updated later.

PR #972, opened 2020-06-16
Add computation for safety metrics to pipeline

liljonnystyle:
## Pull request information

- **Status**: ready to merge

## Description

Using the safety set matrix provided by Marsalis & Jerry, use bilinear interpolation to determine the safety value for every vehicle at every time step. Aggregate measures of safety rate (safe timesteps divided by total timesteps) and max safety score (most dangerous) are computed. The per-timestep data will also be used later to generate safety histograms.


flow/data_pipeline/query.py
```diff
@@ -179,6 +200,41 @@ class QueryStrings(Enum):
...
+            vt.time_step,
+            COALESCE((
+                value_lower_left*(headway_upper-headway)*(rel_speed_upper-leader_rel_speed) + 
+                value_lower_right*(headway-headway_lower)*(rel_speed_upper-leader_rel_speed) + 
+                value_upper_left*(headway_upper-headway)*(leader_rel_speed-rel_speed_lower) + 
+                value_upper_right*(headway-headway_lower)*(leader_rel_speed-rel_speed_lower)
+            ) / ((headway_upper-headway_lower)*(rel_speed_upper-rel_speed_lower)), 200) AS safety_value,
+            vt.source_id
+        FROM fact_vehicle_trace vt
+        LEFT OUTER JOIN fact_safety_matrix sm ON 1 = 1
```
brentgryffindor, line 215: I assume `fact_safety_matrix` is referring to the CSV file you added on S3, is there any updates or maintenance I should perform on it?

flow/data_pipeline/query.py
```diff
@@ -17,6 +20,22 @@
...
+    # "fact_safety_metrics_agg": {
+    #     "leaderboard_chart": [
+    #         "LEADERBOARD_CHART"
+    #     ]
+    # },
+    # "fact_network_throughput_agg": {
+    #     "leaderboard_chart": [
+    #         "LEADERBOARD_CHART"
+    #     ]
+    # }
```
brentgryffindor, line 38: we don't need this for now because `fact_network_fuel_efficiency_agg` should in-theory be run after both  `fact_safety_metrics_agg` and `fact_network_through_agg`. But this is not a hard guarantee since S3 events are indeed asynchronous. I'll be implementing a better way of doing this but in the meantime, we can use the current version first.

PR #976, opened 2020-06-19
count vehicles in domain for leaderboard

liljonnystyle:
## Pull request information

- **Status**: ready to merge

## Description

Add a new query to count the number of vehicles at every time step, so that another plot can be added to the leaderboard.

### DIFF
--- a/flow/data_pipeline/query.py
+++ b/flow/data_pipeline/query.py
@@ -608,18 +608,25 @@ class QueryStrings(Enum):
         ;"""
 
     FACT_VEHICLE_COUNTS_BY_TIME = """
+        WITH counts AS (
+            SELECT
+                vt.source_id,
+                vt.time_step,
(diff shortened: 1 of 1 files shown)


flow/data_pipeline/query.py
```diff
@@ -552,6 +556,22 @@ class QueryStrings(Enum):
         ORDER BY time_seconds_bin ASC
         ;"""
 
+    FACT_VEHICLE_COUNTS_BY_TIME = """
+        SELECT
+            vt.source_id,
+            vt.time_step - FIRST_VALUE(vt.time_step)
+                OVER (PARTITION BY vt.source_id ORDER BY vt.time_step ASC) AS time_step,
```
brentgryffindor, line 563: Presto don't support having window functions in a group by column, I think you'll have to group by first and perform window function on the view here

liljonnystyle: ah good catch

PR #981, opened 2020-06-19
merge custom output and failsafes to master

liljonnystyle:
## Pull request information

- **Status**: ready to merge

## Description

Merge in Brent's custom csv output and Yashar's failsafes updates


AboudyKreidieh:
can we remove all flow/data_pipeline code? we can merge it (if we want to) in a separate PR

flow/core/kernel/vehicle/traci.py
```diff
@@ -1118,3 +1138,69 @@ def get_max_speed(self, veh_id, error=-1001):
...
         self.kernel_api.vehicle.setMaxSpeed(veh_id, max_speed)
+
+    # add for data pipeline
+    def get_accel(self, veh_id):
+        """See parent class."""
+        if "accel" not in self.__vehicles[veh_id]:
+            self.__vehicles[veh_id]["accel"] = None
+        return self.__vehicles[veh_id]["accel"]
+
+    def update_accel_no_noise_no_failsafe(self, veh_id, accel_no_noise_no_failsafe):
```
AboudyKreidieh, line 1149: these are redundant functions, can we replay with something like:
```
def update_accel(self, veh_id, accel, noise=False, failsafe=False):
    ....
```

flow/core/kernel/vehicle/traci.py
```diff
@@ -1118,3 +1138,69 @@ def get_max_speed(self, veh_id, error=-1001):
...
+
+    def update_accel_with_noise_no_failsafe(self, veh_id, accel_with_noise_no_failsafe):
+        """See parent class."""
+        self.__vehicles[veh_id]["accel_with_noise_no_failsafe"] = accel_with_noise_no_failsafe
+
+    def update_accel_with_noise_with_failsafe(self, veh_id, accel_with_noise_with_failsafe):
+        """See parent class."""
+        self.__vehicles[veh_id]["accel_with_noise_with_failsafe"] = accel_with_noise_with_failsafe
+
+    def get_accel_no_noise_no_failsafe(self, veh_id):
```
AboudyKreidieh, line 1165: same here, something like:
```python
def get_accel(self, veh_id, noise=False, failsafe=False):
    ....
```

flow/core/experiment.py
```diff
@@ -153,11 +166,19 @@ def rl_actions(*_):
...
+                if convert_to_csv and self.env.simulator == "traci" and j % 100 == 0:
+                    write_dict_to_csv(trajectory_table_path, extra_info, not j)
+                    extra_info.clear()
+
                 # Compute the results for the custom callables.
                 for (key, lambda_func) in self.custom_callables.items():
                     custom_vals[key].append(lambda_func(self.env))
 
-                if done:
+                if type(done) is dict and done['__all__'] or type(done) is not dict and done:
```
AboudyKreidieh, line 181:
```suggestion
                if type(done) is dict and done['__all__'] or done:
```

flow/core/kernel/vehicle/base.py
```diff
@@ -131,6 +131,23 @@ def remove(self, veh_id):
...
+        Parameters
+        ----------
+        veh_id : str or list of str
+            list of vehicle identifiers
+        acc : float or array_like
+            requested accelerations from the vehicles
+        """
+        raise NotImplementedError
+
+    def apply_acceleration_not_smooth(self, veh_id, acc):
```
AboudyKreidieh, line 145: is this being used in this PR? I can find a function call for it. If it's not, let's leave it for another PR

flow/core/kernel/vehicle/traci.py
```diff
@@ -747,7 +750,7 @@ def _multi_lane_headways(self):
                 for lane in range(max_lanes):
                     edge_dict[edge][lane].sort(key=lambda x: x[1])
 
-        for veh_id in self.get_rl_ids():
+        for veh_id in self.get_ids():
```
AboudyKreidieh, line 753: this could seriously slow down runtime. can we make sure that is not happening before we change it?

flow/core/experiment.py
```diff
@@ -153,11 +166,19 @@ def rl_actions(*_):
                 vel.append(np.mean(self.env.k.vehicle.get_speed(veh_ids)))
                 ret += reward
 
+                # collect additional information for the data pipeline
+                get_extra_info(self.env.k.vehicle, extra_info, veh_ids, source_id, run_id)
```
AboudyKreidieh, line 170: do we always want the extra info? If so, this is not a data-pipeline specific function, and maybe we should move these methods to flow/utils instead

flow/core/kernel/vehicle/base.py
Yasharzf: you can delete this function. I added this when we were testing slowDown vs setSpeed methods. I think we don't need this anymore.

flow/controllers/base_controller.py
```diff
@@ -56,7 +58,18 @@ def __init__(self,
         self.delay = delay
 
         # longitudinal failsafe used by the vehicle
-        self.fail_safe = fail_safe
+        if isinstance(fail_safe, str):
+            self.fail_safe = [fail_safe]
+        elif isinstance(fail_safe, list) or fail_safe is None:
+            self.fail_safe = fail_safe
+        else:
+            print(
```
Yasharzf, line 66: we can replace print with a ValueError exception

liljonnystyle: like this?
```
        try:
            if isinstance(fail_safe, str):
                self.fail_safe = [fail_safe]
            elif isinstance(fail_safe, list) or fail_safe is None:
                self.fail_safe = fail_safe
        except ValueError:
            print(
```

flow/utils/data_pipeline.py
```diff
@@ -0,0 +1,85 @@
...
+    extra_info = pd.DataFrame.from_dict(extra_info)
+    extra_info.set_index(["time", "id"])
+    raw_output = raw_output.merge(extra_info, how="left", left_on=["time", "id"], right_on=["time", "id"])
+
+    # add the partition column
+    # raw_output['partition'] = partition_name
+    raw_output = raw_output.sort_values(by=["time", "id"])
+    output_file_path = data_path[:-4]+"_trajectory.csv"
+    raw_output.to_csv(output_file_path, index=False)
+    return output_file_path
```
brentgryffindor, line 35: this function is kind of obsolete, maybe we shouldn't include it in master

flow/core/experiment.py
```diff
@@ -153,6 +166,14 @@ def rl_actions(*_):
...
+
+                # write to disk every 100 steps
+                if convert_to_csv and self.env.simulator == "traci" and j % 100 == 0:
+                    write_dict_to_csv(trajectory_table_path, extra_info, not j)
```
brentgryffindor, line 174: I'm sorry, can you change this line to `write_dict_to_csv(trajectory_table_path, extra_info, not j and not i)`, please. This is my bad, I forgot to take into consideration of multi-runs. The current code will write a header to the CSV for each run.

flow/core/kernel/vehicle/traci.py
```diff
@@ -87,6 +87,8 @@ def __init__(self,
 
         # old speeds used to compute accelerations
         self.previous_speeds = {}
+        # The time that previous speed is recorded, used to calculate realized_accel
+        self.previous_time = 0
```
brentgryffindor, line 91: This is not actually used. Sorry, I think I just forgot to delete it after I later realized I don't need it.

flow/controllers/base_controller.py
```diff
@@ -235,4 +285,93 @@ def safe_velocity(self, env):
...
+        if this_vel + action * sim_step > edge_speed_limit:
+            if edge_speed_limit > 0:
+                print(
+                    "=====================================\n"
+                    "Speed of vehicle {} is greater than speed limit. Obey "
+                    "speed limit clipping applied.\n"
+                    "=====================================".format(self.veh_id))
+                return (edge_speed_limit - this_vel) / sim_step
+            else:
+                return -this_vel / sim_step
```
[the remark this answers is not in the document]
liljonnystyle: remember this is the acceleration, so it's basically braking to a full stop (subtract all of the velocity in one sim_step)

flow/controllers/base_controller.py
```diff
@@ -90,6 +98,12 @@ def get_action(self, env):
         float
             the modified form of the acceleration
         """
+        # clear the current stored accel_no_noise_no_failsafe of this vehicle None
+        env.k.vehicle.update_accel(self.veh_id, None, noise=False, failsafe=False)
+        env.k.vehicle.update_accel(self.veh_id, None, noise=False, failsafe=True)
+        env.k.vehicle.update_accel(self.veh_id, None, noise=True, failsafe=False)
+        env.k.vehicle.update_accel(self.veh_id, None, noise=True, failsafe=True)
```
nathanlct, line 105: That doesn't seem necessary since all 4 accel values are set afterwards

flow/controllers/base_controller.py
```diff
@@ -107,16 +121,46 @@ def get_action(self, env):
...
+                        env, accel_no_noise_with_failsafe)
+                elif check == 'safe_velocity':
+                    accel_no_noise_with_failsafe = self.get_safe_velocity_action(
+                        env, accel_no_noise_with_failsafe)
+                elif check == 'feasible_accel':
+                    accel_no_noise_with_failsafe = self.get_feasible_action(
+                        accel_no_noise_with_failsafe)
+                elif check == 'obey_speed_limit':
+                    accel_no_noise_with_failsafe = self.get_obey_speed_limit_action(
+                        env, accel_no_noise_with_failsafe)
```
nathanlct, line 142: Maybe do a mapping in `__init__` like

```python
failsafes = { 'instantaneous': self.get_safe_action_instantaneous, 
              'safe_velocity': ... }
```

and then

```python
if self.fail_safe is not None:
    for check in self.fail_safe:
        if check in self.failsafes:
            accel_no_noise_with_failsafe = self.failsafes[check](env, accel_no_noise_with_failsafe)
```

also the `in` could be checked in the constructor so that it's not checked at every iteration:

```python
if isinstance(fail_safe, str):
    self.fail_safe = [fail_safe]
elif isinstance(fail_safe, list) or fail_safe is None:
    self.fail_safe = fail_safe
else:
    raise ValueError("fail_safe should be string or list of strings.\nSet fal_safe to None\n")
    self.fail_safe = None

for check in self.fail_safe:
    if check not in self.failsafes:
        raise ValueError(f'{check} is not a valid failsafe')
```

flow/controllers/base_controller.py
```diff
@@ -56,7 +58,13 @@ def __init__(self,
         self.delay = delay
 
         # longitudinal failsafe used by the vehicle
-        self.fail_safe = fail_safe
+        if isinstance(fail_safe, str):
+            self.fail_safe = [fail_safe]
+        elif isinstance(fail_safe, list) or fail_safe is None:
+            self.fail_safe = fail_safe
+        else:
+            raise ValueError("fail_safe should be string or list of strings.\nSet fal_safe to None\n")
```
nathanlct, line 66:
```suggestion
            raise ValueError("fail_safe should be string or list of strings.\nSetting fail_safe to None\n")
```

flow/controllers/base_controller.py
```diff
@@ -107,16 +121,46 @@ def get_action(self, env):
...
+        if self.fail_safe is not None:
+            for check in self.fail_safe:
+                if check == 'instantaneous':
+                    accel = self.get_safe_action_instantaneous(env, accel)
+                elif check == 'safe_velocity':
+                    accel = self.get_safe_velocity_action(env, accel)
+                elif check == 'feasible_accel':
+                    accel = self.get_feasible_action(accel)
+                elif check == 'obey_speed_limit':
+                    accel = self.get_obey_speed_limit_action(env, accel)
```
nathanlct, line 161: same here, to avoid repetition

flow/controllers/base_controller.py
```diff
@@ -235,4 +285,93 @@ def safe_velocity(self, env):
...
+                "=====================================".format(self.veh_id))
+
+        if action < -self.max_deaccel:
+            action = -self.max_deaccel
+
+            print(
+                "=====================================\n"
+                "Deceleration of vehicle {} is greater than the max "
+                "deceleration. Feasible acceleration clipping applied.\n"
+                "=====================================".format(self.veh_id))
```
nathanlct, line 375: I don't know how often all these failsafe warnings are printed, but if it's every iteration maybe it would be useful to add a printing condition (`self.display_warnings` boolean...)?

flow/core/experiment.py
```diff
@@ -137,10 +140,20 @@ def rl_actions(*_):
         t = time.time()
         times = []
 
+        # data pipeline
+        extra_info = defaultdict(lambda: [])
```
nathanlct, line 144:
```suggestion
        extra_info = defaultdict(list)
```

flow/core/kernel/vehicle/traci.py
```diff
@@ -949,18 +950,22 @@ def _prev_edge_followers(self, veh_id, edge_dict, lane, num_edges):
...
             acc = [acc]
 
         for i, vid in enumerate(veh_ids):
             if acc[i] is not None and vid in self.get_ids():
+                self.__vehicles[vid]["accel"] = acc[i]
                 this_vel = self.get_speed(vid)
                 next_vel = max([this_vel + acc[i] * self.sim_step, 0])
-                self.kernel_api.vehicle.slowDown(vid, next_vel, 1e-3)
+                if smooth:
+                    self.kernel_api.vehicle.slowDown(vid, next_vel, 1e-3)
```
nathanlct, line 966: Is keeping the `smooth` option necessary? I'm not sure a 1e-3s slowdown changes anything when we are dealing with .1 or .5s timesteps

flow/core/kernel/vehicle/traci.py
```diff
@@ -949,18 +950,22 @@ def _prev_edge_followers(self, veh_id, edge_dict, lane, num_edges):
...
+    def apply_acceleration(self, veh_ids, acc, smooth=False):
         """See parent class."""
-        # to hand the case of a single vehicle
+        # to handle the case of a single vehicle
         if type(veh_ids) == str:
             veh_ids = [veh_ids]
             acc = [acc]
 
         for i, vid in enumerate(veh_ids):
             if acc[i] is not None and vid in self.get_ids():
```
nathanlct, line 961: Btw this `in` operates in O(n), maybe it would lead to a little speedup to change `self.__ids` from a list to a set so that it operates in O(1)

flow/controllers/base_controller.py
```diff
@@ -235,4 +285,93 @@ def safe_velocity(self, env):
...
+            the requested action clipped by the speed limit
+        """
+        # check for speed limit
+        this_edge = env.k.vehicle.get_edge(self.veh_id)
+        edge_speed_limit = env.k.network.speed_limit(this_edge)
+
+        this_vel = env.k.vehicle.get_speed(self.veh_id)
+        sim_step = env.sim_step
+
+        if this_vel + action * sim_step > edge_speed_limit:
```
nathanlct, line 328: Would it be useful to create a `env.k.vehicle.get_next_speed(accel)` function? Since this operation is done at several places

flow/core/kernel/vehicle/traci.py
liljonnystyle: @eugenevinitsky do you have any thoughts on this? should we default master to using `setSpeed()`?

flow/core/kernel/vehicle/traci.py
liljonnystyle: @eugenevinitsky I'll leave this to you to decide.

flow/controllers/base_controller.py
liljonnystyle: they need to be set in case the logic is short-circuited (i.e. a few lines below, `if accel is None: return None`)

flow/controllers/base_controller.py
liljonnystyle: I implemented it (here: [3161580](https://github.com/flow-project/flow/pull/981/commits/316158059eaa626726cfe7b5a4e01d8cb607b392)) but then it caused other problems for not having `this_vel` and `sim_step` in scope anymore, so I will keep it as is.

flow/controllers/base_controller.py
nathanlct: right, didn't see that

flow/controllers/base_controller.py
```diff
@@ -235,4 +285,97 @@ def safe_velocity(self, env):
...
+                        "Speed of vehicle {} is greater than speed limit. Obey "
+                        "speed limit clipping applied.\n"
+                        "=====================================".format(self.veh_id))
+                return (edge_speed_limit - this_vel) / sim_step
+            else:
+                return -this_vel / sim_step
+        else:
+            return action
+
+    def get_feasible_action(self, action):
```
nathanlct, line 343: Actually this might fail when calling `accel = failsafe(env, accel)` since there is no `env` arg.. And having an unused arg may annoy pep8

nathanlct: Fixed in bd5845e I suppose

liljonnystyle: oh, just saw your update, cool!

flow/controllers/base_controller.py
```diff
@@ -45,6 +49,7 @@ def __init__(self,
                  car_following_params,
                  delay=0,
                  fail_safe=None,
+                 display_warnings=True,
```
AboudyKreidieh, line 52: (will fix this myself). This needs to be added to every child class otherwise it cannot be modified by anything (e.g. IDMController)

flow/core/kernel/vehicle/traci.py
AboudyKreidieh: I think @nathanlct is right here. Let's keep things as is for now, and if we want to change it later we can do it.

flow/core/kernel/vehicle/traci.py
AboudyKreidieh: This is probably true for many things. We can look into it if it makes things faster (post libsumo). Right now it's probably over-optimization.

PR #983, opened 2020-06-21
Add query to produce max score line in leaderboard

liljonnystyle:
## Pull request information

- **Status**: ready to merge

## Description

Provide a new query that will produce the right data structure to produce the "best score" line in the leaderboard chart correctly.

### DIFF
--- a/flow/data_pipeline/query.py
+++ b/flow/data_pipeline/query.py
@@ -610,7 +610,7 @@ class QueryStrings(Enum):
                     ROWS BETWEEN UNBOUNDED PRECEDING and CURRENT ROW) AS max_score
             FROM leaderboard_chart_agg
             WHERE 1 = 1
-                AND is_baseline = FALSE
+                AND is_baseline = 'False'
         ), prev_max AS (
             SELECT
                 network,
--- a/flow/data_pipeline/query.py
+++ b/flow/data_pipeline/query.py
@@ -44,6 +44,11 @@ tags = {
         "leaderboard_chart_agg": [
             "LEADERBOARD_CHART_AGG"
         ]
+    },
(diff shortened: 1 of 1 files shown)


flow/data_pipeline/query.py
```diff
@@ -594,3 +599,30 @@ class QueryStrings(Enum):
             AND agg.baseline_source_id = baseline.source_id
         ORDER BY agg.submission_date, agg.submission_time ASC
         ;"""
+
+    FACT_TOP_SCORES = """
+        WITH curr_max AS (
+            SELECT
+                network,
+                submission_date,
```
brentgryffindor, line 607: this should be `date AS submission_date` since `leaderboard_chart_agg` don't have a `submission_date` column

flow/data_pipeline/query.py
```diff
@@ -594,3 +599,30 @@ class QueryStrings(Enum):
...
+                1000 * MAX(efficiency_meters_per_joules)
+                    OVER (PARTITION BY network ORDER BY submission_date ASC
```
brentgryffindor, line 609: same here, this should use `date` instead

flow/data_pipeline/query.py
```diff
@@ -594,3 +599,30 @@ class QueryStrings(Enum):
...
+                    ROWS BETWEEN UNBOUNDED PRECEDING and CURRENT ROW) AS max_score
+            FROM leaderboard_chart_agg
+            WHERE 1 = 1
+                AND is_baseline = FALSE
```
brentgryffindor, line 613: `is_baseline` is a string, so this should be `is_baseline = 'False' `

PR #985, opened 2020-06-22
add 210 edgestarts for backwards compatibility

liljonnystyle:
## Pull request information

- **Status**: ready to merge
- **Kind of changes**: bug fix / new feature

## Description

New edgestarts dict is supplied for I-210 subnetwork in case people try to generate the time-space diagram using the emissions csv.

Fast-forwarded PR 989 changes on top, which add greyed regions for warmup period and ghost cells (if applicable). E.g.,
flow_858bc7a51d9f4169872cdc41d20a5cba


liljonnystyle:
@nathanlct Yeah I'm not sure why either. I have rebased master twice and it only seems to have gotten worse.

Also not clear on the -5.1 either, but it seems to be a result of the ghost edge location.

PR #986, opened 2020-06-22
rm duplicate

pengyuan-zhou:


eugenevinitsky:
This build failure is coming from somewhere else. When it's cleared up in master we will merge this.

PR #987, opened 2020-06-25
Query Prereq Check

brentgryffindor:
## Pull request information

- **Status**: ready to merge

## Description

Add the feature to the lambda function that checks and make sure all the prerequisites for running a particular query is met before actually running it.


flow/data_pipeline/data_pipeline.py
```diff
@@ -157,6 +160,39 @@ def update_baseline(s3, baseline_network, baseline_source_id):
...
+    for source_id, completed_queries_list in completed_queries.items():
+        completed_queries_json = json.dumps(completed_queries_list)
+        s3.put_object(Bucket='circles.data.pipeline', Key='lambda_temp/{}'.format(source_id),
+                      Body=completed_queries_json.encode('utf-8'))
+
+
+def get_ready_queries(completed_queries, new_query):
+    """Return queries whose prerequisite queries are completed."""
+    readied_queries = []
+    unfinished_queries = set(prerequisites.keys()) - set(completed_queries)
```
liljonnystyle, line 188: apparently in Python3, `dict.keys()` returns a "set-like view" so it can be used as a set. in that case, perhaps maybe we should make `completed_queries` a set?

brentgryffindor: fixed

PR #1024, opened 2020-11-23
Robustify SUMO restart instance for RLlib at scale

roireshef:
## Pull request information

- **Status**: ready to merge
- **Kind of changes**: bugfix

## Description
Bugfix - When running FLOW-SUMO as a simulation engine within RLlib's training loop, after some substantial amount of steps are accumulated, Flow's Env.reset() will call restart_instance method. When this happens, the process will tend to get stuck and won't manage to restart SUMO successfully to keep training. It appears restarting SUMO process as a daemon when running as a Ray remote process is the root cause, and removing the *setsid* argument solves the problem.


eugenevinitsky:
Okay the travis failure is unrelated to your change. Good catch!

eugenevinitsky (approved):
LGTM nice catch

PR #1026, opened 2020-12-01
Better handling temporary files for running at scale

roireshef:
## Pull request information

- **Status**: ? ready to merge
- **Kind of changes**: ? bug fix

## Description

This PR solves 2 issues:
1. When initializing network kernel with .net.xml filepath, 4 temp files are being created and never removed. This PR added the removal of those files
2. Moving the temp files into the OS default temp directory, rather than in the flow repo. This is mostly critical for running at scale inside docker containers, where the flow repo directory might be massively shared across 100s or 1000s of machines on a shared drive (mapped to a shared folder external to the docker container), and the /debug/cfg folders can eventually get overloaded with files of different machines. The assumption here is that OS default temp folder is not shared across different machines and "lives" only in the context of the docker container, so once the container is killed, any temp files that weren't removed are being wiped out automatically.

**Prior to this change, trying to run flow as the engine for RL experiments at a relatively large scale for some time now, I just found the /debug/cfg full with millions of files "rm *" couldn't even delete because the expansion of "*" made it crash...**


roireshef:
@AboudyKreidieh @eugenevinitsky @kjang96 ?

eugenevinitsky:
Hi Roi, thank you for this! Would you mind running flake8 locally to make sure this conforms to Pep8? We'd really appreciate it so that our tests pass. Otherwise, this looks good!

roireshef:
@eugenevinitsky is it re-running automatically? flake8 pass now.

eugenevinitsky:
Yeah this is erroring due to a change in pip, not your code. Thanks for the PR!

